PR #4, opened 2015-11-07
convert some unwrap() to expect(format!(...))

durka:
`unwrap()` produces panics with no explanation, and on some platforms there are no line numbers in the backtrace even in debug builds (rust-lang/rust#24346). `expect()` with an informative message is a much better user experience.

I didn't convert all instances -- I assume `Regex::new` isn't going to fail since it is used with constants, so I left those. And in a lot of cases the relevant condition is checked -- for instance `if x.contains_key(y) { do_something(x.get(y).unwrap()) }` -- however, these patterns should ideally be refactored to use `match`or other methods to avoid unwrapping.


johannhof:
:+1: thanks

Issue #3, opened 2015-05-18
be multithread-friendly?


blaenk:
I wanted to parse many templates into `Template`s and store them in an `Arc<Mutex<HashMap<String, Template>>>` so that multiple threads could access the template to render various `Context`s, to avoid re-parsing a given template each time it needs to be rendered. However, many things seem to be tangled up in local, non-`'static` references, which means that I can't simply specify `Template<'static>` or `Box<Renderable + Sync + Send + 'static>`. The parser in particular seems to revolve entirely around local references.

Would you guys be open to creating a template type that is "self-contained," so that it is implicitly `Sync + Send + 'static`, to allow us to send and share them between threads? I can do this with the `handlebars` crate, for example, because the [`Template`](http://sunng.info/handlebars-rust/handlebars/struct.Template.html) type is like this. They also have a [`Registry`](http://sunng.info/handlebars-rust/handlebars/struct.Handlebars.html) type that makes it very easy to render a given template "by name," so I simply stick this type in an `Arc<Mutex` after I parse my templates and then various threads can use it to render separate documents.

The only alternative I have now, that I can think of, is to re-parse every template at the time and context that it needs to be rendered, which is a shame.

tak1n:
I'm currently in process of learning rust properly, so no Idea what exactly todo here :smile: .
Feel free to open a PR or wait until I grasp it properly and can change it.

Thx for opening and making some concurrency suggestions

Albibek:
Hi. 
The only thing to do here, as I see it, is requiring Renderable to be Sync+Send. 
Here's an example with tests: https://github.com/Albibek/liquid-rust/commit/a550dc8af51d74abebc86700e6ad3558ebd6843d

johannhof:
Hm that seems reasonable. @Albibek mind making a PR?

epage:
@blaenk  did merging of #63 solve your needs or is there something more?

blaenk:
Although I'm not able to check right now, that does look like it should work. Thanks @Albibek !

PR #8, opened 2016-01-11
Great filters improvement

FerarDuanSednan:
- Allow filter arguments
- Add filters: upcase, minus, replace
- Filter can now work with Value type and not only raw str ( allow filter on arrays, objects, etc )
- changes 100% tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Standard Filters
 - [ ] *pluralize* - return the second word if the input is not 1, otherwise return the first word e.g. {{ 3 | pluralize: 'item', 'items' }} #=> 'items'
 - [ ] *append* - append a string e.g. {{ 'foo' | append:'bar' }} #=> 'foobar'
 - [ ] *slice* - slice a string. Takes an offset and length, e.g. {{ "hello" | slice: -3, 3 }} #=> llo
-- [ ] *minus* - subtraction e.g. {{ 4 | minus:2 }} #=> 2
+- [x] *minus* - subtraction e.g. {{ 4 | minus:2 }} #=> 2
 - [ ] *plus* - addition e.g. {{ '1' | plus:'1' }} #=> 2, {{ 1 | plus:1 }} #=> 2
 - [ ] *times* - multiplication e.g {{ 5 | times:4 }} #=> 20
 - [ ] *divided_by* - integer division e.g. {{ 10 | divided_by:3 }} #=> 3
--- a/README.md
+++ b/README.md
@@ -17,3 +17,42 @@ Now you can use the crate in your code
 ```rust
 extern crate liquid;
 ```
+
+TODO
+---------
+
+Standard Filters
+
+- [ ] *date* - reformat a date (syntax reference)
+- [ ] *capitalize* - capitalize words in the input sentence
+- [ ] *downcase* - convert an input string to lowercase
+- [ ] *upcase* - convert an input string to uppercase
+- [ ] *first* - get the first element of the passed in array
+- [ ] *last* - get the last element of the passed in array
+- [ ] *join* - join elements of the array with certain character between them
+- [ ] *sort* - sort elements of the array
+- [ ] *map* - map/collect an array on a given property
+- [x] *size* - return the size of an array or string
+- [ ] *escape* - escape a string
+- [ ] *escape_once* - returns an escaped version of html without affecting existing escaped entities
+- [ ] *strip_html* - strip html from string
+- [ ] *strip_newlines* - strip all newlines (\n) from string
+- [ ] *newline_to_br* - replace each newline (\n) with html break
+- [ ] *replace* - replace each occurrence e.g. {{ 'foofoo' | replace:'foo','bar' }} #=> 'barbar'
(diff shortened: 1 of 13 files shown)


src/filters.rs
```diff
@@ -1,9 +1,121 @@
...
+    }
+}
+
+pub type FilterResult = Result<String, FilterError>;
+
+pub fn size(input : &Value, _args: &Vec<Value>) -> FilterResult {
+    match input {
+        &Str(ref s)    => Ok(s.len().to_string()),
+        &Array(ref x)  => Ok(x.len().to_string()),
+        &Object(ref x) => Ok(x.len().to_string()),
```
johannhof: you can abbreviate this to `&Str(ref s) | &Array(ref x) | &Object(ref x) => Ok(s.len().to_string())` if I'm not mistaken

README.md
```diff
@@ -17,3 +17,59 @@ Now you can use the crate in your code
 ```rust
 extern crate liquid;
 ```
+Exemple:
```
johannhof: Exemple => Example :)

README.md
```diff
@@ -17,3 +17,59 @@ Now you can use the crate in your code
...
+let mut data = Context::new();
+data.set_val("num", Value::Num(5f32));
+data.set_val("numTwo", Value::Num(6f32));
+
+let output = template.render(&mut data);
+```
+
+Plugins
+--------
+Cache block ( File and Redis ) : https://github.com/FerarDuanSednan/liquid-rust-cache
```
johannhof: :+1:

src/filters.rs
```diff
@@ -1,9 +1,121 @@
...
+            let arg2 = match &args[1] {
+                &Str(ref a) => a, _ => return Err(InvalidArgument(1, "Str expected".to_string()))
+            };
+            Ok(x.replace(arg1, arg2))
+        },
+        _ => Err(InvalidType("String expected".to_string()))
+    }
+}
+
+macro_rules! unit {
```
johannhof: The use of macros is clever, but please wrap everything (line 85-end) in a [tests module](https://doc.rust-lang.org/book/testing.html#the-tests-module). This makes sure it's clear semantically that it's not used anywhere else and helps the compiler.

src/template.rs
```diff
@@ -10,6 +10,9 @@ pub struct Template<'a> {
 impl<'a> Renderable for Template<'a> {
     fn render(&self, context: &mut Context) -> Result<Option<String>> {
         context.filters.insert("size".to_string(), Box::new(size));
+        context.filters.insert("upcase".to_string(), Box::new(upcase));
+        context.filters.insert("minus".to_string(), Box::new(minus));
+        context.filters.insert("replace".to_string(), Box::new(replace));
```
johannhof: We  really need to find a better way to do this than create everything at render eventually, but that's not for this PR :sweat_smile:

johannhof:
This is a great pull request, thanks! I already assumed that transforming strings might no be sufficient for filters. Thanks for changing that. Now implementing more filters will be much easier.

Please correct the minor things I found and we should be ready to go :rocket:

src/filters.rs
FerarDuanSednan: I tried like this:
```rust
&Str(ref x) |&Array(ref x) | &Object(ref x) => Ok(x.len().to_string())
```
But it fail with: _error: mismatched types:_

FerarDuanSednan:
I made the changes.

I think a part of the work I did (commit 53a9655) is related to #3 ( for the use with the iron framework : [ iron-liquid ](https://github.com/FerarDuanSednan/iron-liquid) ).

johannhof:
I suspected as much, very nice! One more thing: I'd love if you could squash this into fewer commits. It doesn't have to be one commit, but please at least get rid of duplicate `Update README.md` and the merge commit. If you need help with rebasing I'd be happy to help you!

FerarDuanSednan:
I made some mistakes during the process but it seems everything is ok now :p

johannhof:
Looks almost perfect! Sorry, there's on more thing I'd like to improve. :snowman:

The first three commits are duplicates of existing commits, but with different hashes (probably happened while rebasing or whatever). Please get rid of them, it would spam the commit history. That shouldn't be too hard I hope, but if you need assistance I'm here :)

FerarDuanSednan:
Done :sparkles:

src/parser.rs
```diff
@@ -76,9 +76,14 @@ fn parse_output<'a>(tokens: &Vec<Token>) -> Result<Box<Renderable + 'a>> {
             return Err(Error::Parser("parse_output: expected a colon".to_owned()));
         }
 
+        iter.next(); // skip colon
+
+        println!("{:?}", tokens);
```
johannhof: Didn't notice that debug print statement before, this is probably not supposed to be there, right? :)

johannhof:
Perfect, thanks! You have a `println!` left in your code I fear :fearful:, but apart from that it LGTM!

FerarDuanSednan:
ups  :scream:, removing this right now  :innocent:

johannhof:
:+1: Thanks for hanging in, great contribution!

PR #15, opened 2016-03-08
Adds for_loop iteration over a range of numbers

tcsc:
The iterated-over range is treated as a created-on-the-fly array of
Value instances, and iterated over the same way as an array stored in
a value. This seemed like the least-awful way of doing using the same
code to render both cases.

Also:
 * Adds the for_loop helper variables as described in "Liquid for
   Designers"
 * Adds the notion of a context scope that can be pushed and popped like
   a stack frame, which is used to automatically clean up the loop
   variables when the loop exits. Implemented as chain of context
   objects, with each deferring to a parent if a variable or filter
   lookup fails.
 * Adds a boolean Value type in an attempt to better mirror Ruby's notion
   truthiness

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,10 +2,15 @@ use std::collections::HashMap;
 use value::Value;
 use filters::Filter;
 
+type ValueMap = HashMap<String, Value>;
+
 #[derive(Default)]
 pub struct Context {
-    values: HashMap<String, Value>,
-    pub filters: HashMap<String, Box<Filter>>,
+    stack: Vec<ValueMap>,
+    globals: ValueMap,
+
+    // Public for backwards compatability
+    pub filters: HashMap<String, Box<Filter>>
 }
 
 impl Context {
@@ -14,8 +19,8 @@ impl Context {
     /// # Examples
     ///
     /// ```
-    /// # use liquid::{Value, Context};
-    /// let mut ctx = Context::new();
+    /// # use liquid::Context;
+    /// let ctx = Context::new();
     /// assert_eq!(ctx.get_val("test"), None);
     /// ```
     pub fn new() -> Context {
@@ -31,14 +36,85 @@ impl Context {
         Context::with_values_and_filters(HashMap::new(), filters)
     }
 
-    pub fn with_values_and_filters(values: HashMap<String, Value>, filters: HashMap<String, Box<Filter>>) -> Context {
+    pub fn with_values_and_filters(values: HashMap<String, Value>,
+                                   filters: HashMap<String, Box<Filter>>) -> Context {
         Context {
-            values: values,
-            filters: filters,
+            stack: vec!(HashMap::new()),
+            globals: values,
+            filters: filters
         }
     }
 
-    /// Gets a value from the rendering context.
+    pub fn add_filter(&mut self, name: &str, filter: Box<Filter>) {
+        self.filters.insert(name.to_owned(), filter);
+    }
+
+    pub fn get_filter<'b>(&'b self, name: &str) -> Option<&'b Box<Filter>> {
+        self.filters.get(name)
+    }
+
+    /// Creates a new variable scope chained to a parent scope.
+    pub fn push_scope(&mut self) {
+        self.stack.push(HashMap::new())
+    }
+
+    /// Removes the topmost stack frame from the local variable stack.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if popping the topmost frame results in an
+    /// empty stack. Given that a context is created with a top-level stack
+    /// frame already in place, empyting the stack should never happen in a
+    /// well-formed program.
+    pub fn pop_scope(&mut self) {
+        if let None = self.stack.pop() {
+            panic!("Pop leaves empty stack")
+        }
+    }
+
+    /// Sets up a new stack frame, executes the supplied function and then
+    /// tears the stack frame down before returning the function's result
+    /// to the caller.
+    ///
+    /// # Examples
+    /// ```
+    /// # use liquid::{Value, Context};
+    /// let mut ctx = Context::new();
+    /// ctx.set_val("test", Value::Num(42f32));
+    /// ctx.with_new_scope(|mut stack_frame| {
+    ///   // stack_frame inherits values from its parent context
+    ///   assert_eq!(stack_frame.get_val("test"), Some(&Value::Num(42f32)));
+    ///
+    ///   // but can (optionally) override them
+    ///   stack_frame.set_local_val("test", Value::Num(3.14f32));
+    ///   assert_eq!(stack_frame.get_val("test"), Some(&Value::Num(3.14f32)));
+    /// });
+    /// // the original value is unchanged once the scope exits
+    /// assert_eq!(ctx.get_val("test"), Some(&Value::Num(42f32)));
+    /// ```
+    pub fn with_new_scope<RvalT, FnT>(&mut self, f: FnT) -> RvalT
+        where FnT : FnOnce(&mut Context) -> RvalT {
+        self.push_scope();
+        let result = f(self);
+        self.pop_scope();
+        result
+    }
+
+    /// Internal part of get_val. Walks the scope stack to try and find the
+    /// reqested variable, and failing that checks the global pool.
+    fn get<'a>(&'a self, name: &str) -> Option<&'a Value> {
+        for frame in self.stack.iter().rev() {
+            if let rval @ Some(_) = frame.get(name) {
+                return rval
+            }
+        }
+        return self.globals.get(name)
+    }
+
+    /// Gets a value from the rendering context. The name value can be a
+    /// dot-separated path to a value. A value will only be returned if
+    /// each link in the chain (excluding the final name) refers to a
+    /// value of type Object.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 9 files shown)


tcsc:
After reading a bit more of "Liquid for Designers" I'm pretty sure the scope idea won't work with the variable assignment examples, as a value created inside a sore loop persists outside it. Still, you have to have some sort of stack, or nested `for_loop`s won't work. This may need more thought.

tcsc:
OK, so I've had a bit of a think about the whole context thing. You might want to hang on to something; this is going to get a bit verbose...

I started from the following premises, gleaned from looking at the "Liquid for ..." documents

1. In general, variables have template scope (i.e. everything is global)
2. for loops, if blocks, etc, can define helper variables which have tag scope, and can be nested.
3. As far as templates are concerned, the variables from (1) and (2) should appear as a unified collection of name/value pairs

I'm also making the assumption that it would be nice to have the stacked contexts automatically clean themselves up as they go out of scope. This implies that there needs to be some sort of stack frame type that represents an individual scope, which can clean up as it is dropped.

The `Context` stack I've implemented here allows (2) _and_ automatic cleanup, but variables assigned inside a loop will be discarded when the loop exits. I don't think the existing design can be easily adapted to having a global pool of variables that is accessible from any level of the stack.

Simply making the parent reference mutable and walking the stack to the ultimate parent for globals is a non-starter; it ends up with the code attempting to create multiple mutable references to the root context (one for the inheritance chain, and one for writing a new value), which is forbidden.

My next attempt was to wrap the globals up in an `Rc` (i.e. `Rc<RefCell<Hashmap<String, Value>>>`) and letting each stack frame borrow a mutable reference when it needs to write a value into the global context. This approach falls down when trying to extract values from the global context - everything expects to retrieve a `&Value`, but the lifespan of a `&Value` extracted from the `RefCell<HashMap>` is only as long as the enclosing scope, meaning it can't be returned to a caller. You _could_ `clone()` the value, but the overhead of copying values on every access seems excessive.

### Where to from here?
One way forward would be to reference count all the values (e.g. something like `Rc<RefCell<Hashmap<String, Rc<Value>>>>`). This would be pretty flexible - it would certainly allow the `Context` to hand out references to `Values` with fewer restrictions, but seems like a pretty drastic change. 

A second other option would be to wear the cost of just copying the Values and be done with it.
A third option is dropping the idea of automatic clean up, and make the tag managing code manually clean up the stack as it goes.

**TL;DR**

Making an automatically managed stack is harder than I thought. Other people's ideas are welcome :-)

johannhof:
Hey, thank you so much for all the work you put into this so far. I'm currently on vacation for another week, but once I get back we'll get this merged and the other problems solved as well! :)

Cheers :heart:

tcsc:
No worries - have a good holiday.

tcsc:
I think I've got it. 

In the end, I decided that the approach of having a scope-managed lifetime was just too hard, so I ditched it altogether. I was inspired by `with_XXX_do` pattern in Haskell and went with manually-managed stack frames, with a convenience function to

1. Set up a stack frame, 
2. Invoke a caller-supplied function,  and 
3. Tear down the stack stack frame before returning the function's result to the caller

This is enough to protect you from a failing `try!()` in a child scope, but not enough to protect you from a panic. My assumption is that if something panics deep in a child context, then all bets are off anyway and  the integrity of your template variables is the least of your worries.

src/context.rs
```diff
@@ -2,10 +2,13 @@ use std::collections::HashMap;
 use value::Value;
 use filters::Filter;
 
+type ValueMap = HashMap<String, Value>;
+
 #[derive(Default)]
 pub struct Context {
-    values: HashMap<String, Value>,
-    pub filters: HashMap<String, Box<Filter>>,
```
johannhof: I'm good with the convenience that `add_filter` offers, but is it really necessary to break backwards compatibility by making `filters` private?

tcsc:
Seem reasonable. Fixed.

johannhof:
Ok great thanks! I'll get to reviewing the rest tomorrow, it looks good to me generally but it's a lot and I'd like to give it a thorough look :)

src/context.rs
```diff
@@ -31,14 +36,85 @@ impl Context {
...
+    pub fn add_filter(&mut self, name: &str, filter: Box<Filter>) {
+        self.filters.insert(name.to_owned(), filter);
+    }
+
+    pub fn get_filter<'b>(&'b self, name: &str) -> Option<&'b Box<Filter>> {
+        self.filters.get(name)
+    }
+
+    /// Creates a new variable scope chained to a parent scope.
+    pub fn push_scope(&mut self) {
```
johannhof: Why are `push_scope` and `pop_scope` public? I thought they were only used internally. I think exposing them would do more harm than good. Am I missing something?

src/context.rs
```diff
@@ -31,14 +36,85 @@ impl Context {
...
+        self.pop_scope();
+        result
+    }
+
+    /// Internal part of get_val. Walks the scope stack to try and find the
+    /// reqested variable, and failing that checks the global pool.
+    fn get<'a>(&'a self, name: &str) -> Option<&'a Value> {
+        for frame in self.stack.iter().rev() {
+            if let rval @ Some(_) = frame.get(name) {
+                return rval
```
johannhof: nit: please end `return` statements with semicolons

src/context.rs
```diff
@@ -31,14 +36,85 @@ impl Context {
...
+            }
+        }
+        return self.globals.get(name)
```
johannhof: see above, or better just remove the `return`

src/context.rs
```diff
@@ -31,14 +36,85 @@ impl Context {
...
+    ///   assert_eq!(stack_frame.get_val("test"), Some(&Value::Num(42f32)));
+    ///
+    ///   // but can (optionally) override them
+    ///   stack_frame.set_local_val("test", Value::Num(3.14f32));
+    ///   assert_eq!(stack_frame.get_val("test"), Some(&Value::Num(3.14f32)));
+    /// });
+    /// // the original value is unchanged once the scope exits
+    /// assert_eq!(ctx.get_val("test"), Some(&Value::Num(42f32)));
+    /// ```
+    pub fn with_new_scope<RvalT, FnT>(&mut self, f: FnT) -> RvalT
```
johannhof: Now, I really like the concept that's behind this function! I think that's an elegant solution.

A small nit: in my terminology of Rust `Context::with_something` would always be used to create a new `Context`. The name confused me for a bit. Maybe we can rename this to e.g. `spawn_scope`? What do you think?

johannhof: Or maybe `spawn` sounds too much like threads. `run_in_scope`?

src/context.rs
```diff
@@ -31,14 +36,85 @@ impl Context {
...
+        self.filters.insert(name.to_owned(), filter);
+    }
+
+    pub fn get_filter<'b>(&'b self, name: &str) -> Option<&'b Box<Filter>> {
+        self.filters.get(name)
+    }
+
+    /// Creates a new variable scope chained to a parent scope.
+    pub fn push_scope(&mut self) {
+        self.stack.push(HashMap::new())
```
johannhof: nit: can you still end this with a semicolon please?

src/context.rs
```diff
@@ -71,16 +163,50 @@ impl Context {
     /// ctx.set_val("test", Value::Num(42f32));
```
johannhof: can you update the example here and move this example to `set_val` above? That'd be amazing! :)

src/value.rs
```diff
@@ -11,8 +11,27 @@ pub enum Value {
     Str(String),
     Object(HashMap<String, Value>),
     Array(Vec<Value>),
+    Bool(bool)
 }
 
+impl Value {
+    pub fn str(val: &str) -> Value {
```
johannhof: I'm not strictly *against* it, but what's the matter with these helper functions that save you from writing `to_owned`? Is this common practice? :sweat_smile:

src/context.rs
tcsc: Done. I originally thought they might be useful outside the protected scope in with_new_scope(), but they're probably not.

src/context.rs
tcsc: done

src/context.rs
tcsc: done

src/context.rs
tcsc: done

src/context.rs
tcsc: Agreed that `with_new_scope` sounds too much like a constructor. I can't think of a better alternative than `run_in_scope()`, so `run_in_scope()` it is.

src/context.rs
tcsc: done

src/value.rs
tcsc: It's common to _me_, at least. I find having to write `to_owned()` for every single string literal to be really, _really_ annoying.  I personally find it clutters up the call site, making it harder to read without adding any extra useful information, so I usually use some sort of function to get rid of it wherever I can. 

Other people's opinions may vary, of course. But that's my 2c.

tcsc:
I've pushed the changes as a separate commit to make it easier to review the delta. Let me know if you'd prefer them squashed.

johannhof:
For two commits it's fine I guess. Thank you so much! I think it's a nice solution and a great write-up.

If you'd like to keep maintaining your code, I can add you as a collaborator to the project. No obligations, just to enable you to review and merge other people's code. :)

tcsc:
Thanks for the offer, but I'm probably going to disappear for a long while very soon: new baby due early May. Probably not the best time to start joining projects ;-)

Issue #11, opened 2016-02-13
Implement missing filters


johannhof:
Feel free to grab any of those and implement! Creating filters is really straightforward, you can find more info [in the README](https://github.com/cobalt-org/liquid-rust#create-your-own-filters) and look at the [existing filters](https://github.com/cobalt-org/liquid-rust/blob/master/src/filters.rs)
- [x] _append_ - append a string e.g. {{ 'foo' | append:'bar' }} #=> 'foobar'
- [x] _capitalize_ - capitalize words in the input sentence
- [x] _ceil_ - 
- [x] _date_ - reformat a date (syntax reference)
- [x] _divided_by_ - integer division e.g. {{ 10 | divided_by:3 }} #=> 3
- [x] _downcase_ - convert an input string to lowercase
- [x] _escape_ - escape a string
- [x] _escape_once_ - returns an escaped version of html without affecting existing escaped entities
- [x] _first_ - get the first element of the passed in array
- [x] _floor_ -
- [x] _join_ - join elements of the array with certain character between them
- [x] _last_ - get the last element of the passed in array
- [x] _map_ - map/collect an array on a given property
- [x] _minus_ - subtraction e.g. {{ 4 | minus:2 }} #=> 2
- [x] _modulo_ - remainder, e.g. {{ 3 | modulo:2 }} #=> 1
- [x] _newline_to_br_ - replace each newline (\n) with html break
- [x] _pluralize_ - return the second word if the input is not 1, otherwise return the first word e.g. {{ 3 | pluralize: 'item', 'items' }} #=> 'items'
- [x] _plus_ - addition e.g. {{ '1' | plus:'1' }} #=> 2, {{ 1 | plus:1 }} #=> 2
- [x] _prepend_ - prepend a string e.g. {{ 'bar' | prepend:'foo' }} #=> 'foobar'
- [x] _remove_first_ - remove the first occurrence e.g. {{ 'barbar' | remove_first:'bar' }} #=> 'bar'
- [x] _remove_ - remove each occurrence e.g. {{ 'foobarfoobar' | remove:'foo' }} #=> 'barbar'
- [x] _replace_first_ - replace the first occurrence e.g. {{ 'barbar' | replace_first:'bar','foo' }} #=> 'foobar'
- [x] _replace_ - replace each occurrence e.g. {{ 'foofoo' | replace:'foo','bar' }} #=> 'barbar'
- [x] _reverse_ - reverse sort the passed in array
[...]

PR #17, opened 2016-03-17
Add Unicode aware Capitalize filter

ligthyear:
refs #11


ligthyear:
@johannhof I wanted to create more filters, but this will be a problem every time if I do them as one PR at a time, as they all modify the same line. Should I just use one branch make one filter per commit and then one bigger PR instead maybe?

johannhof:
@ligthyear sounds like a good solution :+1:

ligthyear:
will superseed this one soon.

PR #18, opened 2016-03-18
Implementing filters for capitalize, downcase, first, last, prepend, append and pluralize

### LINKED ISSUE #11: Implement missing filters

ligthyear:
This implements the a few more filters.

Specifically it adds:

  - [x] capitalize - capitalize words in the input sentence
  - [x] downcase - convert an input string to lowercase
  - [x] first - get the first element of the passed in array
  - [x] last - get the last element of the passed in array
  - [x] prepend - prepend a string e.g. {{ 'bar' | prepend:'foo' }} #=> 'foobar'
  - [x] pluralize - return the second word if the input is not 1, otherwise return the first word e.g. {{ 3 | pluralize: 'item', 'items' }} #=> 'items'
  - [x] append - append a string e.g. {{ 'foo' | append:'bar' }} #=> 'foobar'
slice - slice a string. Takes an offset and length, e.g. {{ "hello" | slice: -3, 3 }} #=> llo

References #11

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -85,6 +85,19 @@ pub fn capitalize(input: &Value, _args: &[Value]) -> FilterResult {
     }
 }
 
+
+pub fn pluralize(input: &Value, args: &[Value]) -> FilterResult {
+
+    if args.len() != 2 {
+        return Err(InvalidArgumentCount(format!("expected 2, {} given", args.len())));
+    }
+    match *input {
+        Num(1f32) => Ok(args[0].clone()),
+        Num(_) => Ok(args[1].clone()),
+        _ => Err(InvalidType("Number expected".to_owned())),
+    }
+}
+
 pub fn minus(input: &Value, args: &[Value]) -> FilterResult {
 
     let num = match *input {
@@ -273,6 +286,15 @@ mod tests {
         assert_eq!(unit!(capitalize, tos!("hello world​")),
                    tos!("Hello World​"));
 
+    }
+
+    #[test]
+    fn unit_pluralize() {
+        assert_eq!(unit!(pluralize, Num(1f32), &[tos!("one"), tos!("many")]),
+                   tos!("one"));
+
+       assert_eq!(unit!(pluralize, Num(2f32), &[tos!("one"), tos!("many")]),
+                  tos!("many"));
     }
 
     #[test]
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -61,6 +61,14 @@ pub fn upcase(input: &Value, _args: &[Value]) -> FilterResult {
 }
 
 
+pub fn downcase(input: &Value, _args: &[Value]) -> FilterResult {
+    match *input {
+        Str(ref s) => Ok(Str(s.to_lowercase())),
+        _ => Err(InvalidType("String expected".to_owned())),
+    }
+}
+
+
 pub fn capitalize(input: &Value, _args: &[Value]) -> FilterResult {
     match *input {
         Str(ref s) => Ok(Str(s.char_indices().fold(String::new(), |word, (_, chr)| {
@@ -197,6 +205,13 @@ mod tests {
         assert_eq!(unit!(upcase, tos!("abc")), tos!("ABC"));
         assert_eq!(unit!(upcase, tos!("Hello World 21")),
                    tos!("HELLO WORLD 21"));
+    }
+
+    #[test]
+    fn unit_downcase() {
+        assert_eq!(unit!(downcase, tos!("Abc")), tos!("abc"));
+        assert_eq!(unit!(downcase, tos!("Hello World 21")),
+                   tos!("hello world 21"));
     }
 
     #[test]
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -183,6 +183,17 @@ pub fn prepend(input: &Value, args: &[Value]) -> FilterResult {
     }
 }
 
+
+pub fn append(input: &Value, args: &[Value]) -> FilterResult {
+    match *input {
+        Str(ref x) => match args.first() {
+                        Some(&Str(ref a)) => Ok(Str(format!("{}{}", x, a))),
+                        _ => return Err(InvalidArgument(0, "Str expected".to_owned())),
+                    },
+        _ => Err(InvalidType("String expected".to_owned())),
+    }
+}
+
 pub fn first(input: &Value, _args: &[Value]) -> FilterResult {
     match *input {
         Str(ref x) => match x.chars().next() {
@@ -323,11 +334,16 @@ mod tests {
                    tos!("foofoo"));
     }
 
-
     #[test]
     fn unit_prepend() {
         assert_eq!(unit!(prepend, tos!("barbar"), &[tos!("foo")]),
                    tos!("foobarbar"));
+    }
+
+    #[test]
+    fn unit_append() {
+        assert_eq!(unit!(append, tos!("sam"), &[tos!("son")]),
+                   tos!("samson"));
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


johannhof:
@ligthyear any updates on this? Should I preliminarily review it already? :)

The travis build is failing because you mis-spelled `divided_by` and it can't import it. :blush:

However, #28 already introduced `divided_by` and `times` so you'll also have to rebase again

ligthyear:
Sorry, @johannhof , I was hoping I had a bit more time to implement a few others, but the last two weeks went crazy instead.

I rebased against master and updated everything to make all tests pass. I guess it is best to review/merge this batch now and do others another time. I'll update the description accordingly.

So, yes, please feel free to review and merge.

johannhof:
@ligthyear don't worry about it, I can totally relate. I'll review it soon! Thanks for contributing

src/filters.rs
```diff
@@ -60,6 +60,46 @@ pub fn upcase(input: &Value, _args: &[Value]) -> FilterResult {
...
+            let next_char = match word.chars().last() {
+                Some(last) =>
+                    if last.is_whitespace() {
+                        chr.to_uppercase().next().unwrap()
+                    } else {
+                        chr
+                    },
+                _ => chr.to_uppercase().next().unwrap(),
+            }.to_string();
+            word + &next_char
```
johannhof: I would've done something like

```
        Str(ref s) => Ok(Str(s.split(' ').map(|word| {
            let (h, t) = word.split_at(1);
            h.to_uppercase() + t
        }).collect::<Vec<String>>().join(" "))),
```
but it seems you're trying to preserve whitespace, which is fine by me. Can I just ask you to add some comments explaining your code, because I really can't wrap my head around it. Also, are the `unwrap` calls guaranteed to be safe?

johannhof: Also, are we sure that the iterator `to_uppercase` returns will always contain only one uppercase letter? afaik `ß` maps to `SS`, for instance.

johannhof:
I basically only have that comment, the rest looks good. We'd also need test coverage for the error cases (passing invalid values to the filters). If you don't have the time for that we can just open a follow-up bug. :)

src/filters.rs
fhartwig: Some random suggestions:

Converting every `char` to a `String` and then appending that to the other string is isn't the best way to do this. You should probably use `String::push` instead. Btw. `word` as the variable name for the accumulator string is a bit misleading, I think.

You're also iterating over the accumulated string to get the last char in every iteration, which means the whole function is O(n^2) in the length of the input string. That's not necessarily a problem, but it would be easy to avoid by just having a boolean that keeps track whether the last char was whitespace. That would make the function O(n) and also make it easier to read.

Finally, you're iterating over `char_indices`, but never actually look at the indices. Is there a reason you couldn't just use `chars` instead?

fhartwig: Regarding the `ß` issue: I don't know about unicode to know how much of a problem this is. `ß` probably won't be a problem because it never occurs at the beginning of a word anyway, But there might be other characters that have the same problem.

ligthyear: Re ß:
ß is the **only letter** that doesn't have any uppercase at all. The work-a-round in technical sense is to often replace it with "SS", which is technically two characters, however that never effects us, as ß also can't be used as the beginning of any word as it should only be used to extend the previous vowel. So, while in a uppercasing of a text this would be relevant (and actually mean your lowercased text is shorter than the uppercased one), in a title-case scenario (the one we are looking at), this isn't an issue in practice. Given, that someone could still feed the ß at the beginning of the word, but as far as I understand this code, this would just lead to the usage of "S" instead of "SS".

ligthyear:
@johannhof See comments in the thread. I can/will take a look this weekend, is that soon enough? also adding some invalid values should be in there ;) .

src/filters.rs
ligthyear: The `unwrap()`s here are safe. We always have an actual character ;) .

fhartwig: > Re Boolean for tracking: indeed, that could be done. That's how I used to do that in python, but never liked it. In case in particular as my otherwise very functional closure has an ugly side effect it constantly depends on.

That doesn't require any side effects, the accumulator variable of the fold can just be `(String, bool)` instead of `String`. Although I have to say that even as a Haskell programmer, I'm not sure that a fold is really better than a loop here.

fhartwig: > Re String::push: good point. Though that means I need to get a mutable handle. I'm sure I can do that in the fold without breaking the borrow checking. I'll try it.

This shouldn't be a problem, you just need to add `mut` to the binding of `word`. I would be surprised if that caused issues with the borrow checker.

johannhof:
Take your time, this PR is not going away  :)

> also adding some invalid values should be in there ;)

huh, where? I can't seem to find it

Thanks!

johannhof:
Going through this again I think it's ok to merge, thank you very much @ligthyear! Feel free to make another PR if you still want to pick up the suggestions in this thread.

Issue #14, opened 2016-02-13
Implement variable assignment


johannhof:
See https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#variable-assignment

tcsc:
I'm working on this at the moment. I've got something close to ready, but it depends on other stuff still in review.

johannhof:
Closed by #19

PR #19, opened 2016-03-20
Adds assignment tag

tcsc:
Addresses Issue #14. Another monster pullreq, sorry. 

In addition to adding the `assign` tag, it also

* Adds context::evaluate() to provide a centralized way of interpreting tokens as values, looking up values from the context as necessary. Modelled on the original Ruby source.
* Moves the definition of `Token` from the lexer into its own module to avoid circular imports.
* Adds some constructor functions for liquid::Error to declutter the call site when returning error values
* Adds Result to the tag parser function signature to allow tag parser to signal failure if necessary
* Adds `parser::expect()` to check that a given token appears in the token stream of a parser when you don't actually care about capturing that value,
* Adds an implicit RHS to the condition parser for `If`, so that it can now parse the "Liquid for designers" examples for `assign` in order to use them as a unit test,
* Pushes the notion of Ruby truthiness for Values into a custom implementaion of PartialEq for `Value`s, to help with the implicit `== true` in an if block.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,6 +1,8 @@
-use std::collections::HashMap;
-use value::Value;
+use error::{Result, Error};
 use filters::Filter;
+use std::collections::HashMap;
+use token::Token::{self, Identifier, StringLiteral, NumberLiteral, BooleanLiteral};
+use value::Value;
 
 type ValueMap = HashMap<String, Value>;
 
@@ -155,6 +157,23 @@ impl Context {
         self.globals.insert(name.to_owned(), val)
     }
 
+    /// Translates a Token to a Value, looking it up in the context if
+    /// necessary
+    pub fn evaluate(&self, t: &Token) -> Result<Option<Value>> {
+        match t {
+            &NumberLiteral(f)     => Ok(Some(Value::Num(f))),
+            &StringLiteral(ref s) => Ok(Some(Value::Str(s.clone()))),
+            &BooleanLiteral(b)    => Ok(Some(Value::Bool(b))),
+            &Identifier(ref id)   => {
+                Ok(self.get_val(id).map(|v| v.clone()))
+            },
+            _ => {
+                let msg = format!("Cannot evaluate {}", t);
+                Err(Error::Other(msg))
+            }
+        }
+    }
+
     /// Sets a value to the rendering context.
     /// Note that it needs to be wrapped in a liquid::Value.
     ///
@@ -223,6 +242,55 @@ mod test {
         // assert that the value has reverted to the old one
         assert_eq!(ctx.get_val("test").unwrap(), &Value::Num(42f32));
         assert_eq!(ctx.get_val("global").unwrap(), &Value::str("some value"));
+    }
+
+    #[test]
+    fn evaluate_handles_string_literals() {
+        use token::Token::StringLiteral;
+
+        let ctx = Context::new();
+        let t = StringLiteral("hello".to_owned());
+        assert_eq!( ctx.evaluate(&t).unwrap(), Some(Value::str("hello")) );
+    }
+
+    #[test]
+    fn evaluate_handles_number_literals() {
+        use token::Token::NumberLiteral;
+
+        let ctx = Context::new();
+        assert_eq!( ctx.evaluate(&NumberLiteral(42f32)).unwrap(),
+                    Some(Value::Num(42f32)) );
+    }
+
+    #[test]
+    fn evaluate_handles_boolean_literals() {
+        use token::Token::BooleanLiteral;
+
+        let ctx = Context::new();
+        assert_eq!(ctx.evaluate(&BooleanLiteral(true)).unwrap(),
+                   Some(Value::Bool(true)) );
+
+        assert_eq!(ctx.evaluate(&BooleanLiteral(false)).unwrap(),
+                   Some(Value::Bool(false)) );
+    }
+
+    #[test]
+    fn evaluate_handles_identifiers() {
+        use token::Token::Identifier;
+
+        let mut ctx = Context::new();
+        ctx.set_val("var0", Value::Num(42f32));
+        assert_eq!(ctx.evaluate(&Identifier("var0".to_owned())).unwrap(),
+                   Some(Value::Num(42f32)));
+        assert_eq!(ctx.evaluate(&Identifier("nope".to_owned())).unwrap(),
+                   None);
+    }
+
+    #[test]
+    fn evaluate_returns_none_on_invalid_token() {
+        use token::Token::DotDot;
+        let ctx = Context::new();
+        assert!(ctx.evaluate(&DotDot).is_err());
     }
 }
 
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -1,69 +1,9 @@
-use self::Token::*;
+use token::Token;
+use token::Token::*;
+use token::ComparisonOperator::*;
 use self::Element::*;
-use self::ComparisonOperator::*;
 use regex::Regex;
 use error::{Error, Result};
-use std::fmt;
-
-#[derive(Clone, Debug, PartialEq)]
-pub enum ComparisonOperator {
-    Equals,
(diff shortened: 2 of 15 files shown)


tcsc:
Bah. Caught by that intermittent timeout when running the skeptic tests. Please re-trigger the build.

johannhof:
Yup, sorry for that, I made #20, I think I have an idea how to fix it.

johannhof:
@tcsc Can you please do another rebase? I've got it "fixed" now. :bowtie:

I'll get around merging or leaving comments today or tomorrow. :)

Thanks!

tcsc:
Rebase done, thanks.

src/value.rs
```diff
@@ -18,20 +18,29 @@ impl Value {
...
+            (&Value::Num(x), &Value::Num(y)) => x == y,
+            (&Value::Str(ref x), &Value::Str(ref y)) => x == y,
+            (&Value::Bool(x), &Value::Bool(y)) => x == y,
+            (&Value::Object(ref x), &Value::Object(ref y)) => x == y,
+            (&Value::Array(ref x), &Value::Array(ref y)) => x == y,
+
+            // encode Ruby truthiness; all values except false and nil
+            // are true, and we don't have a notion of nil
+            (_, &Value::Bool(b)) | (&Value::Bool(b), _) => {
+                println!("****\n {:?}, {:?}", self, b);
```
tcsc: Bah! Stupid `printf` debugging. Will remove.

src/context.rs
```diff
@@ -155,6 +156,23 @@ impl Context {
...
+    ///
+    /// Panics if an unsupported token type is passed via t. Only
+    /// Identifiers and Value Literals are supported.
+    pub fn evaluate(&self, t: &Token) -> Option<Value> {
+        match t {
+            &NumberLiteral(f) => Some(Value::Num(f)),
+            &StringLiteral(ref s) => Some(Value::Str(s.clone())),
+            &BooleanLiteral(b) => Some(Value::Bool(b)),
+            &Identifier(ref id) => self.get_val(id).map(|v| v.clone()),
+            _ => panic!("Unsupported token: {:?}", t)
```
johannhof: I think evaluate is a good idea, but we really shouldn't have this panic since it's used on arbitrary user data and we want to guarantee panic-freeness to programmers using the library. If we have to cover the "unsupported token"  case then we should return a Result instead.

src/context.rs
```diff
@@ -224,5 +242,46 @@ mod test {
...
+    #[test]
+    fn evaluate_handles_identifiers() {
+        use token::Token::Identifier;
+
+        let mut ctx = Context::new();
+        ctx.set_val("var0", Value::Num(42f32));
+        assert_eq!(ctx.evaluate(&Identifier("var0".to_owned())),
+                   Some(Value::Num(42f32)));
+        assert_eq!(ctx.evaluate(&Identifier("nope".to_owned())), None);
+    }
```
johannhof: please add a small test that covers an unsupported token passed to `evaluate`

src/lexer.rs
```diff
@@ -160,6 +101,7 @@ fn granularize(block: &str) -> Result<Vec<Token>> {
             ")" => CloseRound,
             "?" => Question,
             "-" => Dash,
+            "=" => Assignment,
```
johannhof: You will want to add the assigment symbol to the end of [this regex](https://github.com/cobalt-org/liquid-rust/blob/master/src/lexer.rs#L121) so that it gets split properly.

Feel free to also add a test below if you like, I'll take a stab at refactoring these tests soon anyway I think.

src/lib.rs
```diff
@@ -80,7 +81,9 @@ impl Default for ErrorMode {
 ///
 /// ## Minimal Example
 /// ```
-/// # use liquid::{Renderable, LiquidOptions, Context, Error};
+/// # use liquid::{Renderable, LiquidOptions, Context, Error, Tag};
+/// # use liquid::Token;
```
johannhof: why the new use statement and not in the group above?

tcsc: You know, I have no idea. Reverted.

src/value.rs
tcsc: removed

src/context.rs
tcsc: Replaced `panic` with `Result<Option<Value>>`

src/context.rs
tcsc: Done

src/lexer.rs
tcsc: Done. Replaces the existing `==` clause with `={1,2}` which should still preferentially match `==`, but will still match `=`.

src/context.rs
```diff
@@ -155,6 +157,28 @@ impl Context {
         self.globals.insert(name.to_owned(), val)
     }
 
+    /// Translates a Token to a Value, looking it up in the context if
+    /// necessary
+    ///
+    /// # Panics
```
johannhof: I think that's no longer true :)

tests/custom_blocks.rs
```diff
@@ -23,24 +22,27 @@ fn run() {
...
         }).collect();
-        Box::new(Multiply{numbers: numbers}) as Box<Renderable>
+        Ok(Box::new(Multiply{numbers: numbers}))
     }
 
-    let mut tags = HashMap::new();
-    tags.insert("multiply".to_string(), Box::new(multiply_tag) as Box<Tag>);
-
-    let options = LiquidOptions {
+    let mut options = LiquidOptions {
```
johannhof: nit: since all fields are set to default here, you might as well just set `options` to `Default::default()`

src/tags/for_block.rs
```diff
@@ -118,13 +108,9 @@ fn parser_err<T>(expected: &str, actual: Option<&Token>) -> Result<T> {
...
 fn int_attr<'a>(args: &mut Iter<'a, Token>) -> Result<Option<usize>> {
-    match args.next() {
-        Some(&Colon) => (),
-        x => return parser_err(":", x)
-    };
-
+    try!(expect(args, Colon));
     match args.next() {
         Some(&NumberLiteral(ref n)) => Ok(Some(*n as usize)),
         x => return parser_err("number", x)
```
johannhof: Can't you replace `parser_err` with `Error::parser` now? (and delete the `parser_err fn?)

src/value.rs
```diff
@@ -18,20 +18,28 @@ impl Value {
...
-        }
-    }
+impl PartialEq<Value> for Value {
+    fn eq(&self, other: &Value) -> bool {
+        match (self, other) {
+            (&Value::Num(x), &Value::Num(y)) => x == y,
+            (&Value::Str(ref x), &Value::Str(ref y)) => x == y,
+            (&Value::Bool(x), &Value::Bool(y)) => x == y,
+            (&Value::Object(ref x), &Value::Object(ref y)) => x == y,
+            (&Value::Array(ref x), &Value::Array(ref y)) => x == y,
```
johannhof: can't you aggregate these match arms using the pipe (`|`) operator?

johannhof:
Ok some minor things left and then we're good to merge!

src/value.rs
tcsc: I din't think so, the `x`s and `y`s are all different types, so I don't think the variable binding will work. I'll give it a go tonight, though, and see.

johannhof: Oh that might be, I didn't try it out. It's really not important.

src/context.rs
tcsc: indeed

tests/custom_blocks.rs
tcsc: done

src/tags/for_block.rs
tcsc: done

src/value.rs
tcsc: Yep. Tried it and it fails due to the mismatched types.

johannhof:
Great work again! Thank you very much! :+1:

Issue #21, opened 2016-03-23
Implement missing tags


tak1n:
https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#tags

- [x] assign - Assigns some value to a variable
- [x] capture - Block tag that captures text into a variable
- [x] case - Block tag, its the standard case...when block
- [x] comment - Block tag, comments out the text in the block
- [x] cycle - Cycle is usually used within a loop to alternate between values, like colors or DOM classes.
- [x] for - For loop
- [x] break - Exits a for loop
- [x] continue Skips the remaining code in the current for loop and continues with the next loop
- [x] if - Standard if/else block
- [x] include - Includes another template; useful for partials
- [x] raw - temporarily disable tag processing to avoid syntax conflicts.
- [x] unless - Mirror of if statement

tcsc:
Any suggestions of which would be the most useful to attack? I've got a few ideas on how to approach `cycle`, and figuring out a sane way to do `break` or `continue` would be ... interesting.

PS: I also have an implementation of `capture` that will be ready for review once some other stuff passes review.

tak1n:
@tcsc the most useful one is the one which was requested on cobalt.rs: https://github.com/cobalt-org/cobalt.rs/issues/61.
But we should implement all of them in future.

Maybe you can work together with @LucioFranco who is also interested in implementing the include tag.

Regarding to your pull request, I still have to get comfortable with liquid-rust therefore @johannhof is in charge for merging and reviewing pull requests for liquid-rust. :)

tcsc:
No worries. A naïve implementation of `include` should be pretty straightforward. 

It does present the opportunity for a pretty large security hole, though. I mean what if someone were to `{% include /root/.ssh/id_rsa %}.`... Oops. Might be worth seeing if shopify does anything special. Will have a peek.

tcsc:
...and I didn't mean the bit about the pullreq to be naggy. It was just a heads up, so someone  didn't write the same code twice. Apologies if it came off whiney. :-)

tak1n:
@tcsc yep we should see how liquid from shopify is handling that cases.

Also it didn't came off whiney or naggy at all, just wanted to give you some feedback as of why I'm not dealing with your PR's.

We appreciate you're contributions thanks :smile:

johannhof:
> Apologies if it came off whiney

Didn't seem that way at all. :)

[...]

PR #22, opened 2016-03-24
Adds capture block

tcsc:
Implements the capture block for #21.

Finally, a sane pull request of reasonable size! :-)


johannhof:
No objections :+1:

PR #24, opened 2016-03-24
added Cargo.lock to .gitignore

LucioFranco:
This fixes an issue where when you run `cargo build` in a clean working directory, and it becomes dirty with an untracked file that is `Cargo.lock`. So I just added `Cargo.lock` to the `.gitignore` similar to how it is done in `cobalt-org/cobalt.rs`.


johannhof:
Thanks :+1:

PR #25, opened 2016-03-25
Add Include Tag

LucioFranco:
This PR adds the `include` tag. Example:  `{% include path/to/file.liquid %}`.

This is related to #21 and [#61](https://github.com/cobalt-org/cobalt.rs/issues/61).

This is still a very basic implementation that uses the path that the process was run from for relative paths. This might pose issues where you build from a different directory that you run the cobalt command from. For most projects this should work just fine.

**Edit**: I also added a simple file cache that way we do not need to hit the fs if we include the samefile.


src/tags/include_tag.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[cfg(test)]
+mod test {
+    use context::Context;
+    use Renderable;
+    use parse;
+
+    #[test]
+    fn include_tag() {
+        let text = "{% include tests/fixtures/additional/default.liquid %}";
```
tcsc: I would try not to depend on an external file being in the right place at the right time for a unit test - I've been bitten by that too often. If you can't fake the file access (e.g. read from a file-like memory buffer or something), and in this case I think the test really _does_ need to hit the filesystem, I usually:

1. Create a temp directory at the start of the test,
2. Create the file with the content I want to test in that directory
3. Point whatever code that needs the file at the temp file
4. Destroy the temp dir on test exit

Have a peek at the[ `tempdir`](https://doc.rust-lang.org/tempdir/tempdir/index.html) crate, which gives you an RAII scope-managed temp directory, which would probably do a lot of the heavy lifting for you.

(...but I'll defer to @johannhof if he thinks I'm being too pedantic)

LucioFranco: @tcsc The file I added is in the git repo. As long as the `cargo test` command is run from the root of the git repo then everything should work. If it doesn't then there is an issue with the code/repo.

src/tags/include_tag.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+pub fn include_tag(_tag_name: &str,
+                   arguments: &[Token],
+                   _options: &LiquidOptions)
+                   -> Result<Box<Renderable>> {
+    let mut args = arguments.iter();
+
+    match args.next() {
+        Some(&Token::Identifier(ref path)) => Ok(Box::new(Include { partial_path: path.clone() })),
```
tcsc: Should the parser load and parse the template here, and then save the parsed template in the `Include` struct?

That way

1. if someone caches the outer template then you only have to hit the filesystem once, rather than every time the template is rendered, and
2. you don't get the renderer raising parser errors, so things are a bit more predicable.

LucioFranco: @tcsc the one thing that made me do it in the renderer is because the actual `Error::renderer` fit better than `Error::parse`. Loading the file from the fs is not really a parse error. 

The idea of caching is interesting but would require adding some stuff to `context` I believe.

tcsc: Even without changing the `Context`, it's still possible for calling code to construct a template instance and render it over & over again. 

The original shopify implementation in Ruby actually caches the `include`d fragments so they can be used multiple times in a single "host" template, which would definitely require adding some sort of extra machinery in the `Context` struct.

src/tags/include_tag.rs
johannhof: In this case I'd be ok with it. I usually go by: for read operations, it's ok to use fixture files, if you want to write stuff, mock your filesystem. If we run into any problems that could be solved by tempdir we can still switch later on.

Cobalt is a big culprit of writing test stuff to the local repo and I had this `TODO: switch to tempdir` in my head forever now.

tl;dr I'm fine with @LucioFranco's solution

src/tags/include_tag.rs
johannhof: You need include the file, lex and parse it at compile time, since we don't want to do the lexing and parsing at render time. While this might not make a big difference to Cobalt, I've heard of people who want to use liquid on web servers, where render is called tons of times per minute.

"File not found" is a parser error, as strange as it sounds. Rust's libsyntax also does not successfully parse a file with incorrect `mod` statements. It makes sense if you consider that the `include`d file might also throw parser errors that we want to catch in the parsing step.

You don't have to use `Error::parse(expected, actual)`, you can make your own `Error::Parser` with a custom message.

You should basically copy what [liquid::parse](https://github.com/cobalt-org/liquid-rust/blob/master/src/lib.rs#L149) already does, without registering the tags and blocks and just passing the options through to the parser.

Then you can save the resulting `Template` in your struct and call `template.render(context)` at render time. :)

LucioFranco:
@johannhof @tcsc please take a look. I have moved the fs operations to the parse stage and the partial is now parsed as well.

src/tags/include_tag.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    #[test]
+    fn include_tag() {
+        let text = "{% include tests/fixtures/additional/default.liquid %}";
+        let template = parse(text, Default::default()).unwrap();
+
+        let mut context = Context::new();
+        assert_eq!(template.render(&mut context).unwrap(),
+                   Some("hello, world!\n".to_owned()));
+    }
```
johannhof: can you add a test that checks there's an error if the file was not found?

src/tags/include_tag.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+impl Renderable for Include {
+    fn render(&self, mut context: &mut Context) -> Result<Option<String>> {
+        self.partial.render(&mut context)
+    }
+}
+
+fn parse_partial(path: &str, options: &LiquidOptions) -> Result<Template> {
+    let path = Path::new(&path);
```
johannhof: this line should not be necessary, because `&str` can convert to `Path`

src/tags/include_tag.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    // check if file exists
+    if !path.exists() {
+        return Error::parser_msg(&format!("{:?} does not exist", path));
+    }
```
johannhof: you can leave this out I think, if the path doesn't exist `open` will throw a similar error

src/tags/include_tag.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    let mut file = match File::open(path) {
+        Ok(file) => file,
+        Err(e) => return Error::parser_msg(&format!("[std::io::Error] {}", e)),
+    };
+
+    let mut content = String::new();
+    if let Err(e) = file.read_to_string(&mut content) {
+        return Error::parser_msg(&format!("[std::io::Error] {}", e));
+    }
```
johannhof: please go to `error.rs` and add a `From` implementation for `io:Error`, just copy and paste it from Cobalt: https://github.com/cobalt-org/cobalt.rs/blob/master/src/error.rs#L21

You can then make your code much shorter, e.g.:
```rust
let mut file = try!(File::open(path));
let mut content = String::new();
try!(file.read_to_string(&mut content));
```

src/error.rs
```diff
@@ -20,8 +20,11 @@ pub enum Error {
 
 impl Error {
     pub fn parser<T>(expected: &str, actual: Option<&Token>) -> Result<T> {
-        Err(Error::Parser(
-            format!("Expected {}, found {:?}", expected, actual)))
+        Err(Error::Parser(format!("Expected {}, found {:?}", expected, actual)))
+    }
+
+    pub fn parser_msg<T>(msg: &str) -> Result<T> {
+        Err(Error::Parser(msg.to_owned()))
```
johannhof: if you see below comments you can remove this again, we don't need it anymore :)

tests/fixtures/input/example.txt
```diff
@@ -1 +1,2 @@
 {{'whooo' | size}}{%comment%}What happens{%endcomment%} {%if num < numTwo%}wat{%else%}wot{%endif%} {%if num > numTwo%}wat{%else%}wot{%endif%}
+{% include tests/fixtures/additional/default.liquid %}
```
johannhof: If you want you can instead add another set of tests under  `tests/fixtures/{input,output}/include.txt` that check things like if included files get parsed and evaluated properly.

src/tags/include_tag.rs
LucioFranco: @johannhof You are right but the reason I did not use that error is because the error is just `Io(Error { repr: Os { code: 2, message: "No such file or directory" } })'` where the error message that I use has the file name in it. `Parser("\"file_does_not_exsist.liquid\" does not exist")`. To me this seems like the better option that will notify the user that the included file is not a file without having to run the command with `--trace` to see which file could not be found.

johannhof: I see your point! That's fine with me then. :+1:

johannhof: But can you then please turn this into `Error::Parser(format!("{:?} does not exist", path));` and remove the `parser_msg` function?

johannhof:
@LucioFranco one more thing and we should be good to go (see my last comment). Your build is failing because of changes in Clippy on nightly. I fixed it on master, you'll have to rebase.

Thanks!

src/tags/include_tag.rs
LucioFranco: done

tests/fixtures/input/example.txt
LucioFranco: done

LucioFranco:
@johannhof I added all the things you suggested and its passing so I think this PR is ready!

johannhof:
Yay! Good job! Thank you for your contribution! :+1:

PR #26, opened 2016-03-25
Adds `break` and `continue` tags.

tcsc:
Adds the notion of an "interrupt" - a signal in the rendering context that stops all rendering and returns the rendered content back up the call stack. When the stack unwind hits a for_loop, it resets the
interrupt and continues processing as directed - either breaking out of the loop or continuing on to the next iteration.

The `break` and `continue` tags are then implemented on top of this interrupt construct. This is pretty much how the shopify implementation handles the same problem.

Also adds support for nested tags in the parser (e.g. nested `for_loop`, `if`, etc). Prior to this, the parser would truncate the children at the inner `end` tag.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -4,12 +4,22 @@ use std::collections::HashMap;
 use token::Token::{self, Identifier, StringLiteral, NumberLiteral, BooleanLiteral};
 use value::Value;
 
+#[derive(Clone)]
+pub enum Interrupt { Continue, Break }
+
 type ValueMap = HashMap<String, Value>;
 
 #[derive(Default)]
 pub struct Context {
     stack: Vec<ValueMap>,
     globals: ValueMap,
+
+    /// The current interrupt state. The interrupt state is used by
+    /// the `break` and `continue` tags to halt template rendering
+    /// at a given point and unwind the `render` call stack until
+    /// it reaches an enclosing `for_loop`. At that point the interrupt
+    /// is cleared, and the `for_loop` carries on processing as directed.
+    interrupt: Option<Interrupt>,
 
     // Public for backwards compatability
     pub filters: HashMap<String, Box<Filter>>
@@ -42,6 +52,7 @@ impl Context {
                                    filters: HashMap<String, Box<Filter>>) -> Context {
         Context {
             stack: vec!(HashMap::new()),
+            interrupt: None,
             globals: values,
             filters: filters
         }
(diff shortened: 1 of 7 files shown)


src/parser.rs
```diff
@@ -108,12 +109,23 @@ fn parse_tag(iter: &mut Iter<Element>,
...
+                        },
+                        ref n if n == &end_tag && nesting_depth > 0 => {
+                            nesting_depth -= 1;
+                        },
+                        ref n if n == &end_tag && nesting_depth == 0 => {
+                            break
+                        },
+                        _ => {}
+                    }
+                };
```
johannhof: can you quickly elaborate (and maybe write a short comment) why you need to collect the nesting_depth here?

src/template.rs
```diff
@@ -20,6 +20,14 @@ impl Renderable for Template {
             if let Some(ref x) = try!(el.render(context)) {
                 buf = buf + x;
             }
+
+            // Did the last element we processed set an interrupt? If so, we
+            // need to abandon the rest of our child elements and just
+            // return what we've got. This is usually in respinse to a
```
johannhof: typo: respinse -> response

src/context.rs
```diff
@@ -11,6 +14,13 @@ pub struct Context {
     stack: Vec<ValueMap>,
     globals: ValueMap,
 
+    /// The current interrupt state. The interrupt state is used by
+    /// the `break` and `continue` tags to halt template rendering
+    /// at a given point and unwind the `render` call stack until
+    /// it reaches an enclosing `for_loop`. At that point the interrupt
+    /// is cleared, and the `for_loop` carries on procesisng as directed.
```
johannhof: typo: procesisng -> processing :)

src/context.rs
```diff
@@ -55,6 +66,22 @@ impl Context {
         self.filters.get(name)
     }
 
+    pub fn interrupted(&self) -> bool {
+        self.interrupt.is_some()
+    }
+
+    /// Sets the interrupt state. Any previous state is oblierated.
```
johannhof: typo: obliterated

johannhof:
Mmh, this one is a bit tricky. I'm usually not a big fan of changing things all around the code to support a single tag. But in this case there really seems to be no better alternative. I left a few minor comments :)

src/parser.rs
tcsc: done

src/template.rs
tcsc: done

src/context.rs
tcsc: done

src/context.rs
tcsc: done

src/parser.rs
johannhof: ah, clever indeed

johannhof:
:+1: great, thanks!

Issue #23, opened 2016-03-24
Implement unless block


johannhof:
...which is basically just the opposite of `if`. You can find the `if` code here: https://github.com/cobalt-org/liquid-rust/blob/master/src/tags/if_block.rs

This should be fairly straightforward, however, please do not fully copy-paste the if code into a new file, we should find a way to share some code between the two tags.

johannhof:
Closed by #27

PR #27, opened 2016-03-29
Adds `unless` and `elsif` support

tcsc:
Adds `unless` support (see #23) to the if_block.rs module. Also adds
support for `elsif` blocks in an `if`, which are parsed into
recursive else blocks.

Also adds a sub-block aware splitter function that can be used to
split a tag body on any of a set of delimiters, where the split is
guaranteed to happen at the top levek of nesting (e.g, attempting
to split the tokens inside an `if` block on an `else` is guaranteed
to split the token stream only on an  `else` block at the same
nesting level as the `if` (as opposed to, say, the `else` block of
a `for_loop` nested inside the body of the `if`).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,8 +42,8 @@ extern crate regex;
 
 use std::collections::HashMap;
 use lexer::Element;
-use tags::{assign_tag, include_tag, break_tag, continue_tag, comment_block, raw_block, for_block,
-           if_block, capture_block};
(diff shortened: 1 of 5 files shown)


johannhof:
I merged your #26, wanna rebase? :)

tcsc:
Rebased.

johannhof:
Sorry, took me a while to look at this but I wanted to be sure I understand the BlockSplit concept. I think I can live with this solution, thanks for the effort!

PR #28, opened 2016-03-29
Add most of the remaining math filters

bfrog:
* times
* divided_by
* ceil
* floor
* round

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -84,6 +84,50 @@ pub fn plus(input: &Value, args: &[Value]) -> FilterResult {
     }
 }
 
+pub fn times(input: &Value, args: &[Value]) -> FilterResult {
+
+    let num = match *input {
+        Num(n) => n,
+        _ => return Err(InvalidType("Num expected".to_owned())),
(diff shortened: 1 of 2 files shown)


johannhof:
Hey, thanks for contributing! Your PR is looking good, just two minor things:

- you'll have to rebase on master, sorry for that :)
- I'd love if you could add assertions for the error returning paths as well, doesn't have to be fancy (e.g. call "ceil" with a string and check it returns an error)

Thank you!

bfrog:
I spent few thinking about a way to make a macro for testing error returns without adding a bunch of fluff, might take me awhile. This was rebased otherwise though with the success cases tested so far.

johannhof:
Can't you just do 
```rust
assert!(ceil(Bool(true), &[]).is_err());
```
? :)

bfrog:
Right, I was for whatever reason thinking testing the error type and string would also be good, but I guess just showing its an error scenario would be good enough

johannhof:
Great job, thank you! :+1:

PR #30, opened 2016-04-06
Add support for relative paths

LucioFranco:
This allows a user of `liquid-rust` to do the following:
```
let options = LiquidOptions {
            blocks: Default::default(),
            tags: Default::default(),
            relative_path: "tests/fixtures/input".to_owned(),
            error_mode: Default::default(),
        };
let text = "{% include example.txt %}";
let template = parse(text, options).unwrap();

let mut context = Context::new();
assert_eq!(template.render(&mut context).unwrap(),
                   Some("5 wot wot\n".to_owned()));
```

This will allow `cobalt.rs` to supply the build path to liquid so the include paths can be more readable and less repetative.

### DIFF
--- a/src/tags/include_tag.rs
+++ b/src/tags/include_tag.rs
@@ -10,7 +10,7 @@ use error::{Result, Error};
 
 use std::fs::File;
 use std::io::Read;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 struct Include {
     partial: Template,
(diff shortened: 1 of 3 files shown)


src/tags/include_tag.rs
```diff
@@ -59,11 +60,21 @@ mod test {
...
     use parse;
     use error::Error;
+    use LiquidOptions;
+
+    fn options () -> LiquidOptions {
+        LiquidOptions {
+            blocks: Default::default(),
+            tags: Default::default(),
+            relative_path: "tests/fixtures/input".to_owned(),
+            error_mode: Default::default(),
```
johannhof: you can also just write 

```rust
LiquidOptions {
  relative_path: "tests/fixtures/input".to_owned(),
  ..Default::default()
}
```

johannhof:
I like the original Liquid name of `filesystem`/`file_system` a bit better than relative_path, can we change that? Also, I feel like the API would be more clear if the type would be `Option<Path>`, what do you think?

src/tags/include_tag.rs
LucioFranco: I did not know that! Thanks will add it.

LucioFranco:
All done! Should be ready!

johannhof:
Good job! :+1:

PR #34, opened 2016-04-15
Moved include tag to String Literal

### LINKED ISSUE #31: Include tag proper template names
In `Shopify/liquid` the syntax for the include tag is different from my implementation.

```
  # Simply include another template:
  #
  #   {% include 'product' %}
  #
  # Include a template with a local variable:
  #
  #   {% include 'product' with products[0] %}
  #
```

- [x] With String literal
- [ ] With variable

We should implement the Include tag similar. This issue is to keep track.

LucioFranco:
Fixes #31. Still need to figure out variable part and if we want to include it.

### DIFF
--- a/src/tags/include_tag.rs
+++ b/src/tags/include_tag.rs
@@ -47,9 +47,15 @@ pub fn include_tag(_tag_name: &str,
     let mut args = arguments.iter();
 
     let path = match args.next() {
-        Some(&Token::Identifier(ref path)) => path,
+        Some(&Token::StringLiteral(ref path)) => path,
+
+        Some(&Token::Identifier(_)) => {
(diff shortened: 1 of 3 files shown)


johannhof:
:+1: , but why can't we have both, for flexibility?

LucioFranco:
@johannhof What do you mean by both?

johannhof:
We could support both identifiers and string literals for specifying the path, or is that weird because it's actually a variable?

LucioFranco:
I don't think we should do it because a identifier is a variable usually. Maybe we could add a nice error message when a user tries an identifier and they meant a String Literal.

johannhof:
That sounds nice, wanna make a good-first-bug for it?

PR #39, opened 2016-06-08
Pass on intermediate filter result (fixes #38).

### LINKED ISSUE #38: Chaining filters ignores intermediate results
There appears to be an issue if you chain more than one filter together as shown in this sample program
```
extern crate liquid;
use liquid::{Renderable, Context, Value};

fn main() {
    println!("Run cargo test instead!");
}

#[test]
fn test_uppercase() {
    let template = liquid::parse("{{ s | upcase }}", Default::default()).unwrap();

    let mut context = Context::new();
    context.set_val("s", Value::Str(String::from("foofoo")));

    assert_eq!(output.unwrap(), Some("FOOFOO".to_string()));
}
#[test]
fn test_replace() {
    let template = liquid::parse("{{ s | replace:'foo','bar' }}", Default::default()).unwrap();

    let mut context = Context::new();
    context.set_val("s", Value::Str(String::from("foofoo")));

    assert_eq!(output.unwrap(), Some("barbar".to_string()));
}

#[test]
fn test_both() {
    let template = liquid::parse("{{ s | replace:'foo','bar' | upcase }}", Default::default()).unwrap();

    let mut context = Context::new();
    context.set_val("s", Value::Str(String::from("foofoo")));

    assert_eq!(output.unwrap(), Some("BARBAR".to_string()));
}
```
Here is the result:
```
% cargo test
     Running target/debug/loops-43680f964937768f

running 3 tests
test test_uppercase ... ok
test test_replace ... ok
test test_both ... FAILED

failures:

---- test_both stdout ----
thread 'test_both' panicked at 'assertion failed: `(left == right)` (left: `Some("FOOFOO")`, right: `Some("BARBAR")`)', src/main.rs:37

failures:
    test_both

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured
```
The intermediate filter result isn't passed as input to the next filter in the chain.  Should have gotten `BARBAR` but instead for `FOOFOO`.  It seems like the `replace` filter wasn't called, but I added some debugging `println!` and is _IS_ being called, but its result isn't input to the next block in the chain.

Quick scan of the code seems that the issue is around [in this area](https://github.com/cobalt-org/liquid-rust/blob/master/src/output.rs#L43-56)

johannhof:
Costs us another .clone() because we have to give ownership to the entry variable.


johannhof:
@LucioFranco can you review this? I don't think I'm responsible for the coverage decrease.

LucioFranco:
@johannhof try a rebase. Added the same fix I added for `cobalt.rs`.

PR #42, opened 2016-06-13
Add date filter

kstep:
A date filter implementation for #11 from cobalt-org/cobalt.rs#115.

---
This change is [Reviewable](https://reviewable.io/reviews/cobalt-org/liquid-rust/42)

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -8,7 +8,7 @@ use chrono::DateTime;
 
 use self::FilterError::*;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub enum FilterError {
     InvalidType(String),
     InvalidArgumentCount(String),
@@ -333,6 +333,15 @@ mod tests {
         }};
     }
 
+    macro_rules! failed {
+        ( $a:ident, $b:expr ) => {{
+            failed!($a, $b, &[])
+        }};
+        ( $a:ident, $b:expr, $c:expr ) => {{
+            $a(&$b, $c).unwrap_err()
+        }};
+    }
+
     macro_rules! tos {
         ( $a:expr ) => {{
             Str($a.to_owned())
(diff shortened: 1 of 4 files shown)


src/filters.rs
```diff
@@ -232,6 +234,25 @@ pub fn last(input: &Value, _args: &[Value]) -> FilterResult {
     }
 }
 
+pub fn date(input: &Value, args: &[Value]) -> FilterResult {
+    if args.len() != 1 {
+        return Err(FilterError::InvalidArgumentCount(format!("expected 1, {} given", args.len())));
+    }
+
+    let date = match input {
+        &Value::Str(ref s) => try!(DateTime::parse_from_str(&s, "%d %B %Y %H:%M:%S %z")
```
johannhof: So the original liquid has a `to_date` function that I think is a good idea to implement for us as well https://github.com/Shopify/liquid/blob/master/lib/liquid/utils.rb#L63

However, we don't have to do it in this PR. I'll make a follow-up issue.

src/filters.rs
```diff
@@ -232,6 +234,25 @@ pub fn last(input: &Value, _args: &[Value]) -> FilterResult {
...
+                                   .map_err(|e| FilterError::InvalidType(format!("Invalid date format: {}", e)))),
+        _ => return Err(FilterError::InvalidType("String expected".to_owned())),
+    };
+
+    let format = match args[0] {
+        Value::Str(ref s) => s,
+        _ => return Err(InvalidArgument(0, "Str expected".to_owned())),
+    };
+
+    Ok(Value::Str(format!("{}", date.format(format))))
```
johannhof: wouldn't it be more efficient to just call `date.format(format).to_string()`?

src/filters.rs
```diff
@@ -383,4 +404,9 @@ mod tests {
         assert_eq!(unit!(last, Array(vec![tos!("test"), tos!("last")])), tos!("last"));
         assert_eq!(unit!(last, Array(vec![])), tos!(""));
     }
+
+    #[test]
+    fn unit_date() {
+        assert_eq!(unit!(date, tos!("13 Jun 2016 02:30:00 +0300"), &[tos!("%Y-%m-%d")]), tos!("2016-06-13"));
```
johannhof: Could you add tests that pass in wrong values for the three error conditions in your function? There are some (unfortunately not enough) examples in the tests above.

johannhof:
Sorry, I just got around looking at this. Looks great apart from the minor comments! You'll also have to do a rebase from master.

Thanks again!

johannhof:
Oh, I didn't notice the update. Thanks, looks great now! 👍 👍 👍

PR #46, opened 2016-06-15
implement join filter

bacoboy:
needed to pull in the `itertools` crate since there doesn't
seem to be a good way to do a string join in Rust w/o tons of
code.  I'm open to suggestions for doing this w/o the crate
addition.

Also not a huge fan of the Array validation/unwrapping implementation.
Also open to suggestions.

But it works!

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -269,6 +269,34 @@ pub fn split(input: &Value, args: &[Value]) -> FilterResult {
     }
 }
 
+pub fn join(input: &Value, args: &[Value]) -> FilterResult {
+    // Make sure there is only 1 argument to join
+    if args.len() != 1 {
+        return Err(InvalidArgumentCount(format!("expected 1, {} given", args.len())));
+    }
+
+    match *input {
+        Array(ref array) => {
+            // use ToStr to stringify the values in case they aren't strings...
+            let mut strings_to_join = array.iter().map(|x| x.to_string());
+            // the input is in fact an Array of Strings
+            match args.first() {  // Check the first (and only) argument
+                Some(&Str(ref join_string)) => {
+                    // The join string argument is in fact a String
+                    let mut result = strings_to_join.next().unwrap_or(String::new());
+                    for string in strings_to_join {
+                        result.push_str(&join_string);
+                        result.push_str(&string);
+                    }
+                    Ok(Str(result))
+                }
+                _ => Err(InvalidArgument(0, "expected String argument as join".to_owned())),
+            }
+        }
+        _ => Err(InvalidType("Array of Strings expected".to_owned())),
+    }
+}
+
 #[cfg(test)]
 mod tests {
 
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -15,6 +15,7 @@ build = "build.rs"
 [dependencies]
 regex = "0.1"
 lazy_static = "0.1.15"
+itertools = "0.4.15"
```
johannhof: please remove this dependency if you're not using it anymore :)

src/filters.rs
```diff
@@ -269,6 +271,38 @@ pub fn split(input: &Value, args: &[Value]) -> FilterResult {
...
+    match *input {
+        Array(ref array) => {
+            // Not a fan of this, but need to unwrap the Strs and also check for errors
+            let mut strings_to_join = Vec::new();
+            for value in array {
+                match *value {
+                    Str(ref s) => {
+                        strings_to_join.push(s);
+                    }
+                    _ => return Err(InvalidType("Array of Strings expected".to_owned())),
```
johannhof: I wonder if we really have to expect an array of strings or just implicitly convert all values with `value.to_string()`

src/filters.rs
```diff
@@ -435,4 +469,13 @@ mod tests {
...
                    Array(vec![tos!("a"), tos!("b")]));
     }
+
+    // unit! macro doesn't like an Array as an input so will
+    // skip unit test in favor of more complete test.
+    // Not sure how to adjust the macro to support Array...
+    // #[test]
+    // fn unit_join() {
+    //     assert_eq!(unit!(join, Array(vec![tos!("a"), tos!("b"),
+    //                tos!("c")]), &[tos!(",")], tos!("a,b,c"));
```
johannhof: you don't have to use `unit!`, you can simply call `join` normally and check the result. It's a bit more verbose, but does the job.

johannhof:
Looks good, thanks!

> Also not a huge fan of the Array validation/unwrapping implementation.

Yeah I can understand that, but it's the cleanest way to properly check that all input is valid. That's the overhead you get from supporting actual strong types, not just "stringly" typing in template languages. But I believe it's worth it.

Cargo.toml
bacoboy: This is what implements the `join()` on line 296.
As I said, I wasn't sure of a better way to do String joins.  Most every reference to the question led me to this crate...

src/filters.rs
bacoboy: That is a very good point.  Update soon...

src/filters.rs
bacoboy: updated.

Cargo.toml
johannhof: Sorry, I was traveling. You could solve it like this:
```rust
let mut result = strings_to_join[0].to_string();
for string in strings_to_join.iter().skip(1).map(|x| x.as_str()) {
    result.push_str(join_string);
    result.push_str(string);
}
Ok(Str(result))
```
However, you can shorten the code above (where you declare `strings_to_join`) to simply:

```rust
let mut strings_to_join = array.iter().map(|x| x.to_string());
```

which enables us to simplify the loop below a lot:

```rust
let mut result = strings_to_join.next().unwrap_or(String::new());
for string in strings_to_join {
         result.push_str(&join_string);
         result.push_str(&string);
}
Ok(Str(result))
```
And that allows us to avoid a lot of type conversion and solves it in a single iteration of `array`.

bacoboy: I see you added an `as_str()` to `Value` which helps a lot.  Updated.

src/filters.rs
```diff
@@ -269,6 +269,34 @@ pub fn split(input: &Value, args: &[Value]) -> FilterResult {
...
+            // use ToStr to stringify the values in case they aren't strings...
+            let mut strings_to_join = array.iter().map(|x| x.to_string());
+            // the input is in fact an Array of Strings
+            match args.first() {  // Check the first (and only) argument
+                Some(&Str(ref join_string)) => {
+                    // The join string argument is in fact a String
+                    let mut result = strings_to_join.next().unwrap_or(String::new());
+                    for string in strings_to_join {
+                        result.push_str(&join_string);
+                            result.push_str(&string);
```
johannhof: nit: indent :)

bacoboy: thought I ran format.. dang it...

bacoboy:
Looks like something changed with the nightly outside the scope of this change...
https://travis-ci.org/cobalt-org/liquid-rust/jobs/139298535

johannhof:
Yeah don't worry, Clippy is broken. Good to merge!

PR #50, opened 2016-07-08
Allow for variables to be passed as filter args

### LINKED ISSUE #49: Error using variables as filter arguments
When trying to use a variable as an argument to a filter I get an error from the parser.

Consider this example:

```rust
let input = "{{ \"test\" | prepend: myvar }}";
let parsed = match liquid::parse(input, Default::default()) {
    Ok(t) => t,
    Err(e) => { panic!("{}", e) }
};
```

Results in:

```
thread '<main>' panicked at 'Parsing error: Expected a comma or a pipe, found myvar'
```

gchp:
This commit adds support for passing variables as arguments to filter.
It changes the FilterPrototype type to contain a list of VarOrVal
instances rather than just Value instances.

When filters are being applied, each VarOrVal instance is "unwrapped"
before being given to the filter invocation. For VarOrVal::Val this
means just extracting the inner Value. For VarOrVal::Var it means
looking up the variable value.

This fixes #49


src/parser.rs
```diff
@@ -84,18 +84,31 @@ pub fn parse_output(tokens: &[Token]) -> Result<Output> {
...
+                        return Error::parser("a comma or a pipe", Some(&Token::Identifier(v.clone())));
+                    }
+                    args.push(VarOrVal::Var(Variable::new(v)))
+                }
                 x => {
                     return Error::parser("a comma or a pipe", Some(x));
                 }
             }
+            comma_previous = false;
+            first = false;
```
gchp: I don't really like this piece.

Essentially, a test was failing because instead of raising an error for a missing comma or pipe, the parser was eating the next identifier, producing the wrong result.

To get around this, I'm keeping track of whether or not this is the first time through the list, and whether the previous token was a comma. If a comma was not directly before the new Identifier, and its not the first time through the loop, then we raise an error. This fixes the test.

A better way to do this would probably be to look ahead after parsing the Identifier and check if the next token is either a comma or a pipe. What do you think? Any thoughts / suggestions?

Let me know if you need me to clarify. If you want to see the failing test, just revert the second commit in this PR.

gchp:
Will update to fix the fmt errors. Let me know what you think re the above comment!

src/parser.rs
johannhof: Mmh the original code seems generally problematic since we just swallow any value without looking at commas in between. As far as I understand the Liquid "spec" there should always be a comma or a pipe after any filter argument. 

The smartest thing to do might be to add a `peek` after the `match` that checks if we have either a comma or a pipe after our argument and fail if we don't. If there is a comma we can consume it and `continue` and if there's a Pipe we can just `break`. Then you'll obviously need to remove the Comma match clause.

Do you think that'll work?

src/output.rs
```diff
@@ -59,7 +59,20 @@ impl Output {
         for filter in &self.filters {
             let f = try!(context.get_filter(&filter.name)
                 .ok_or(Error::Render(format!("Filter {} not implemented", &filter.name))));
-            entry = try!(f(&entry, &filter.arguments));
+
+            let mut arguments = Vec::new();
+            for arg in filter.arguments.iter() {
```
johannhof: I think the `iter` can be inferred here

src/output.rs
```diff
@@ -59,7 +59,20 @@ impl Output {
...
+                match arg {
+                    &VarOrVal::Var(ref x) => {
+                        let val = context.get_val(&*x.name()).cloned().unwrap_or(Value::Str("".to_owned()));
```
johannhof: instead of defaulting to a `Value::Str` you should throw a render error saying that the variable is not defined

johannhof:
Hey, thanks so much for the PR! I left some comments.

Cheers

src/output.rs
gchp: Was getting errors when I removed the iter. Changed it to loop over `&filter.arguments` instead.

src/parser.rs
gchp: I think so - will give it a shot and update!

gchp: Done!

johannhof:
Looks great! 👍 

One thing before we can merge: you accidentally committed `src/parser.rs.bk` (we should just add these to the .gitignore). Please remove it! :)

gchp:
Whoops, sorry! Didn't realise I added that. Will update in the morning when I get back to my laptop.

gchp:
Ok - updated!

johannhof:
Great job! Thanks! 👍

PR #51, opened 2016-07-11
Only add default filters if not already set

gchp:
This allows for users to override the default filters.


johannhof:
Sounds reasonable! Thanks!

PR #52, opened 2016-07-11
Allow arguments to include tag to be non-quoted

gchp:
This allows for non-string-literals to be given as arguments to the
include tag. This is supported by the Ruby library, and I think it makes
sense to support here too.


johannhof:
I agree!

PR #53, opened 2016-07-11
Allow dashes in Identifiers

gchp:
This allows things like:

    {% include my-file.html %}

Previously this example would have produced the tokens:

    Identifier("include")
    Identifier("my")
    Dash
    Identifier("file.html")

After this change, the token stream is:

    Identifier("include")
    Identifier("my-file.html")

This is again something that is supported by the ruby library, which
would be great to support here.


johannhof:
I'd say in this case people could simply use a string in the include tag, but since this doesn't seem to cause any conflicts I see no harm in supporting it, either. Thanks for the PR!

PR #54, opened 2016-07-12
Prevent panic using limit in for block

gchp:
Previously using a "limit" in a for block greater than the length of the
iterable would cause the renderer to panic.

Instead we now check the length of the iterable and if the limit is
greater than the length we pick the whole iterable, otherwise (as
before) slice to the end of the iterable.


johannhof:
Great, thanks!

PR #58, opened 2016-08-14
Implement liquid::parse_file(...) method

### LINKED ISSUE #57: Add a from_file function
Would be nice to allow loading template files directly from the file system.

ElijahCaine:
Hey I really like your project and I'm glad to be able to contribute.

If you have any questions about what / why I did something please ask. I'm fairly new to Rust so there are a few '*decisions*' which were really 'I *think* this is how to accomplish that thing I want to do but who knows'.

Implements basic expected functionality:
- Takes as arument a filepath and `LiquidOptions`.
- Returns `Result<Template>`
Includes basics tests (extends `fixtures.rs`).
Includes doc-string documentation.
fixes #57

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,7 +70,9 @@ use lexer::Element;
 use tags::{assign_tag, cycle_tag, include_tag, break_tag, continue_tag, comment_block, raw_block,
            for_block, if_block, unless_block, capture_block, case_block};
 use std::default::Default;
-use std::path::PathBuf;
+use std::fs::File;
+use std::io::prelude::Read;
+use std::path::{PathBuf, Path};
 use error::Result;
 
 pub use value::Value;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -214,3 +216,36 @@ pub fn parse(text: &str, options: LiquidOptions) -> Result<Template> {
...
+/// use liquid::{Renderable, LiquidOptions, Context};
+///
+/// let template = liquid::parse_file("/path/to/template.liquid",
+///                                    LiquidOptions::default()).unwrap();
+/// let mut data = Context::new();
+/// data.set_val("num", Value::Num(4f32));
+/// let output = template.render(&mut data);
+/// assert_eq!(output.unwrap(), Some("Liquid! 2".to_string()));
+/// ```
+pub fn parse_file(fp: &str, options: LiquidOptions) -> Result<Template> {
```
johannhof: You can use [AsRef](https://doc.rust-lang.org/std/convert/trait.AsRef.html) to enable automatic conversion of supported types to `Path`, which allows for a more flexible API. Basically just copy the function signature of [File::open](https://doc.rust-lang.org/std/fs/struct.File.html#method.open).

tests/fixtures.rs
```diff
@@ -51,3 +66,37 @@ pub fn include_with_context() {
 
     compare("include_with_context", &mut context);
 }
+
+#[test]
+pub fn graceful_fail_when_nonexistent_file() {
+    // TODO: Write this test
```
johannhof: I would name this test `error_on_nonexistent_file`, since I wouldn't say returning a failed `Result` is particularly graceful. It's debatable, I guess. Anyway, I think my name fits the behavior a bit better :)

This test shouldn't be hard to write, just pass a bogus path to `parse_file` and assert that the result is an error.

src/lib.rs
```diff
@@ -70,6 +70,8 @@ use lexer::Element;
 use tags::{assign_tag, cycle_tag, include_tag, break_tag, continue_tag, comment_block, raw_block,
            for_block, if_block, unless_block, capture_block, case_block};
 use std::default::Default;
+use std::fs::File;
+use std::io::prelude::*;
```
johannhof: Can't you reduce this to `use std::io::prelude::Read;`?

johannhof:
Hey, thanks so much for contributing! This patch looks great!

A note on the tests: I'm not sure `fixtures.rs` is the best place for testing the `parse_file` method. Why don't you make a new test file `tests/parse_file.rs` where you check that `parse_file` works correctly. Testing a couple of fixtures and checking the error cases (nonexistent file) should be enough :)

Cheers!

ElijahCaine:
@johannhof I took your feedback and made the suggested changes. I had no idea that AsRef design pattern was a thing in Rust, thank's for that.

I also merged my commits for a cleaner git history, hope that's okay.

Let me know if you have any other notes.

johannhof:
Looks great, thanks again!

In case you're looking for more stuff to do, check out the list of missing filters (#11) :wink:

PR #59, opened 2016-08-17
Fix parse_file(...) docs.

ElijahCaine:
I was reading the docs for parse_file and I realized they were formatted totally incorrectly. I basically ended up adding some whitespace and now it looks a lot better and I tested it locally with `cargo doc`.

Sorry for *yet another* Pull Request, couldn't leave this un-fixed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -217,27 +217,30 @@ pub fn parse(text: &str, options: LiquidOptions) -> Result<Template> {
     parser::parse(&tokens, &options).map(Template::new)
 }
 
-/// Parse a liquid template from a file, returning Result<Template, io::Error>
+/// Parse a liquid template from a file, returning a `Result<Template, Error>`.
 /// # Examples
 ///
 /// ## Minimal Template
 ///
-/// template.liquid:
-/// ```
-/// "Liquid! {{num | minus: 2}}"
+/// `template.txt`:
+///
+/// ```text
+/// "Liquid {{data}}"
 /// ```
 ///
 /// Your rust code:
-/// ```
-/// use liquid::{Renderable, LiquidOptions, Context};
 ///
-/// let template = liquid::parse_file("/path/to/template.liquid",
-///                                    LiquidOptions::default()).unwrap();
+/// ```rust,no_run
+/// use liquid::{Renderable, LiquidOptions, Context, Value};
+///
+/// let template = liquid::parse_file("path/to/template.txt",
+///                                   LiquidOptions::default()).unwrap();
 /// let mut data = Context::new();
-/// data.set_val("num", Value::Num(4f32));
+/// data.set_val("data", Value::Num(4f32));
 /// let output = template.render(&mut data);
-/// assert_eq!(output.unwrap(), Some("Liquid! 2".to_string()));
+/// assert_eq!(output.unwrap(), Some("Liquid 4\n".to_string()));
 /// ```
+///
 pub fn parse_file<P: AsRef<Path>>(fp: P, options: LiquidOptions) -> Result<Template> {
     let mut options = options;
     options.register_known_blocks();


src/lib.rs
```diff
@@ -217,18 +217,20 @@ pub fn parse(text: &str, options: LiquidOptions) -> Result<Template> {
     parser::parse(&tokens, &options).map(Template::new)
 }
 
-/// Parse a liquid template from a file, returning Result<Template, io::Error>
+/// Parse a liquid template from a file, returning a Result (Template or io Error).
```
johannhof: you don't have to remove the angle brackets, you can simply put it in backticks: 
```
`Result<Template, io::Error>`
```

johannhof:
oh, ha, I didn't notice this wasn't actually tested. Thanks for taking another look.

Your tests are failing because you don't import `liquid::Value`. Should be easy to fix. :)

src/lib.rs
```diff
@@ -217,18 +217,20 @@ pub fn parse(text: &str, options: LiquidOptions) -> Result<Template> {
...
 /// # Examples
 ///
 /// ## Minimal Template
 ///
-/// template.liquid:
+/// `template.liquid`:
+///
 /// ```
```
johannhof: You need to add an `ignore` after the backticks here to avoid [this failure](https://travis-ci.org/cobalt-org/liquid-rust/jobs/152865914#L541)

ElijahCaine:
@johannhof  I fixed the docs, the checks pass now. What do you think?

src/lib.rs
```diff
@@ -217,18 +217,20 @@ pub fn parse(text: &str, options: LiquidOptions) -> Result<Template> {
...
+/// `template.liquid`:
+///
+/// ```text
 /// "Liquid! {{num | minus: 2}}"
 /// ```
 ///
 /// Your rust code:
-/// ```
+///
+/// ```rust,ignore
```
johannhof: Don't `ignore` here, this should run as a test.

src/lib.rs
```diff
@@ -217,18 +217,20 @@ pub fn parse(text: &str, options: LiquidOptions) -> Result<Template> {
...
 /// use liquid::{Renderable, LiquidOptions, Context};
```
johannhof: You still need to use `Value` as well here :)

ElijahCaine:
@johannhof Sorry about the delay. I pushed a fix so the tests actually get run. The workaround is to parse one of the test's fixture inputs. If you think there's a cleaner way to get that done I'm up for it, but as it stands the docs run the test and they are pretty good at explaining what needs to be done.

Let me know what you think.

johannhof:
@ElijahCaine no worries, thanks for the update! In fact, I'm really sorry for forgetting that this test needs an actual file to pass.

Your solution is fine, but doesn't really help the readability of the docs IMO. I preferred your previous version. A compromise could be to add the `no_run` attribute to the test instead (see https://doc.rust-lang.org/book/documentation.html#running-documentation-tests). That would compile the test but not run it, which should be enough.

Thanks!

ElijahCaine:
@johannhof I agree that the fix didn't help the readability so I added the `no_run` attribute to the code-block and fixed things up a bit. Let me know how what you think.

src/lib.rs
```diff
@@ -217,27 +217,29 @@ pub fn parse(text: &str, options: LiquidOptions) -> Result<Template> {
...
+/// ```rust,no_run
+/// use liquid::{Renderable, LiquidOptions, Context, Value};
+///
+/// let template = liquid::parse_file("path/to/template.txt", LiquidOptions::default()).unwrap();
 /// let mut data = Context::new();
-/// data.set_val("num", Value::Num(4f32));
+/// data.set_val("data", Value::Num(4f32));
 /// let output = template.render(&mut data);
-/// assert_eq!(output.unwrap(), Some("Liquid! 2".to_string()));
+/// assert_eq!(output.unwrap(), Some("Liquid 4\n".to_string()));
```
johannhof: why the newline delimiter after the 4?

src/lib.rs
```diff
@@ -217,27 +217,29 @@ pub fn parse(text: &str, options: LiquidOptions) -> Result<Template> {
...
 /// Your rust code:
-/// ```
-/// use liquid::{Renderable, LiquidOptions, Context};
 ///
-/// let template = liquid::parse_file("/path/to/template.liquid",
-///                                    LiquidOptions::default()).unwrap();
+/// ```rust,no_run
+/// use liquid::{Renderable, LiquidOptions, Context, Value};
+///
+/// let template = liquid::parse_file("path/to/template.txt", LiquidOptions::default()).unwrap();
```
johannhof: please keep the indentation style here :)

johannhof:
Thanks! Just two minor nits before we can merge!

src/lib.rs
ElijahCaine: The tests wouldn't pass when we it used a real file because it ended in a newline character. I can get rid of it since I agree that it's confusing (even if it is accurate).

ElijahCaine: I added a note under the code-block that I think will be a good compromise.

> Note: Since most editors place a newline character (\n) at the end of lines, the test only plasses when a newline character is included in the expected output.

ElijahCaine:
@johannhof I made those changes. I didn't remove the newline but I did explain why it was there. Let me know if there's anything else I should tweak.

johannhof:
@ElijahCaine oh hey, sorry I missed that. I actually think you can remove the explanation. Not sure why I was wondering about the newline. I guess it's best to be accurate, even if the test won't run.

Sorry about the confusion. We can merge after that :)

ElijahCaine:
@johannhof Ready when you are (and the CI checks turn green I assume :smile:)

johannhof:
Yaaaay 👍

PR #60, opened 2016-10-25
Value traits and sort filter

Albibek:
Hi.
- Implemented PartialEq correctly and completely, as it was asked in TODO 
- Added Eq and Ord for Value
- Made `sort` filter with test

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -299,6 +299,23 @@ pub fn join(input: &Value, args: &[Value]) -> FilterResult {
     }
 }
 
+pub fn sort(input: &Value, args: &[Value]) -> FilterResult {
+    if args.len() > 0 {
+        return Err(FilterError::InvalidArgumentCount(format!("expected no arguments, {} given",
+                                                             args.len())));
+    }
+    match input {
+        &Value::Array(ref array) => {
+            let mut sorted = array.clone();
+            sorted.sort();
+            Ok(Value::Array(sorted))
+        }
+        _ => Err(FilterError::InvalidType("Array argument expected".to_owned())),
+
+    }
+}
+
+
 pub fn date(input: &Value, args: &[Value]) -> FilterResult {
     if args.len() != 1 {
         return Err(FilterError::InvalidArgumentCount(format!("expected 1, {} given", args.len())));
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -301,8 +301,7 @@ pub fn join(input: &Value, args: &[Value]) -> FilterResult {
 
 pub fn sort(input: &Value, args: &[Value]) -> FilterResult {
     if args.len() > 0 {
-        return Err(FilterError::InvalidArgumentCount(format!("expected no arguments, {} given",
-                                                             args.len())));
+        return Err(InvalidArgumentCount(format!("expected no arguments, {} given", args.len())));
     }
     match input {
         &Value::Array(ref array) => {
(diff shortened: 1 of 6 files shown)


src/filters.rs
```diff
@@ -299,6 +299,23 @@ pub fn join(input: &Value, args: &[Value]) -> FilterResult {
     }
 }
 
+pub fn sort(input: &Value, args: &[Value]) -> FilterResult {
+    if args.len() > 0 {
+        return Err(FilterError::InvalidArgumentCount(format!("expected no arguments, {} given",
```
johannhof: nit: you can leave out the `FilterError::` here

src/filters.rs
```diff
@@ -299,6 +299,23 @@ pub fn join(input: &Value, args: &[Value]) -> FilterResult {
...
+                                                             args.len())));
+    }
+    match input {
+        &Value::Array(ref array) => {
+            let mut sorted = array.clone();
+            sorted.sort();
+            Ok(Value::Array(sorted))
+        }
+        _ => Err(FilterError::InvalidType("Array argument expected".to_owned())),
```
johannhof: nit: you can leave out the `FilterError::` here

tests/filters.rs
```diff
@@ -253,13 +252,13 @@ pub fn split_no_comma() {
 
 #[test]
 // Split on 1 string and re-join on another
-pub fn split_then_join() {
```
johannhof: Please create a new test instead of overwriting this one :)

src/value.rs
```diff
@@ -61,48 +61,24 @@ impl PartialEq<Value> for Value {
     }
 }
 
-// TODO implement for object and array
-// TODO clean this up
+impl Eq for Value {}
```
johannhof: I suppose we can't use `derive` instead?

src/value.rs
```diff
@@ -61,48 +61,24 @@ impl PartialEq<Value> for Value {
...
-            _ => false,
-        }
-    }
-    fn ge(&self, other: &Value) -> bool {
-        match (self, other) {
-            (&Value::Num(x), &Value::Num(y)) => x.ge(&y),
-            (&Value::Str(ref x), &Value::Str(ref y)) => x.ge(y),
-            (&Value::Bool(x), &Value::Bool(y)) => x.ge(&y),
-            _ => false,
-        }
```
johannhof: Nice!
This reminds me: Can you please add a couple of assertions to the `number_equality` test that verify that `<`, `>`, `<=` and `>=` comparisons work correctly? Thanks!

src/value.rs
```diff
@@ -61,48 +61,24 @@ impl PartialEq<Value> for Value {
...
+}
+
+impl Ord for Value {
+    fn cmp(&self, other: &Value) -> Ordering {
+        self.partial_cmp(other).unwrap_or(Ordering::Equal)
```
johannhof: Do we really want to fall back to equal here?

Albibek:
About ordering. That's my bad of thinking about this only in terms of sorting. I think it would me more correct to remove the Ord impl and move the same like that to sort filter using `sort_by` instead of `sort`. It should give the intented level of comparison while sorting the array where all types are same while not touching the global inconsistence as it was with my approach.
We could even support partial equality of floats adding some loginc and returning error from sort filter, but I'm not sure we really need such level of correctness.
What do you think?

johannhof:
> I think it would me more correct to remove the Ord impl and move the same like that to sort filter using sort_by instead of sort. It should give the intented level of comparison while sorting the array where all types are same while not touching the global inconsistence as it was with my approach.

That sounds great to me.

> We could even support partial equality of floats adding some loginc and returning error from sort filter, but I'm not sure we really need such level of correctness.

Not sure if I understand you correctly, but it sounds unnecessary :)

Thanks!

Albibek:
I meant that floats, for example, don't impl Ord, because NaNs are not orderable. Anyways that is rare case, having no meaning in templates. See last commit for the fix.

johannhof:
I see, thanks for the explanation. This looks good to me now. :)

PR #61, opened 2016-11-02
Slice filter implementation

Albibek:
Implemented slice filter as of #11. 
Also some formatting was added automatically, that seems to be correct.


johannhof:
Great PR and well tested. I have zero complaints. 👍 

Thanks!

PR #64, opened 2016-11-19
Sort filters

vvv:
- Template::render(): Sort filter names alphabetically.
- src/filters.rs: Leave a single blank line between function definitions.


johannhof:
Looks good, thanks!

PR #65, opened 2016-11-19
Add `modulo' filter

vvv:
+ src/filters.rs (sort): Use is_empty() instead of comparing len() with 0.
+ tests/filters.rs: Separate function definitions with a single blank line.


johannhof:
Looks great, thanks!

PR #69, opened 2016-11-20
[refactoring] Use a function to check the number of arguments

vvv:
src/filters.rs (check_args_len): New function.


johannhof:
👍

PR #70, opened 2016-11-20
Add remove_first, remove, replace_first filters, strip_html and truncatewords

kubahorak:
Check out the new filters!

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -451,14 +451,16 @@ pub fn remove(input: &Value, args: &[Value]) -> FilterResult {
 }
 
 pub fn strip_html(input: &Value, _args: &[Value]) -> FilterResult {
-    // regexps taken from https://git.io/vXbgS
-    let matchers = [Regex::new(r"(?is)<script.*?</script>").unwrap(),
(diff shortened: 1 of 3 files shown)


src/filters.rs
```diff
@@ -398,6 +426,47 @@ pub fn escape(input: &Value, args: &[Value]) -> FilterResult {
...
+        _ => Err(InvalidType("String expected".to_owned())),
+    }
+}
+
+pub fn strip_html(input: &Value, _args: &[Value]) -> FilterResult {
+    // regexps taken from https://git.io/vXbgS
+    let matchers = [Regex::new(r"(?is)<script.*?</script>").unwrap(),
+                    Regex::new(r"(?is)<style.*?</style>").unwrap(),
+                    Regex::new(r"(?is)<!--.*?-->").unwrap(),
+                    Regex::new(r"(?is)<.*?>").unwrap()];
```
johannhof: You could use lazy_static to initialize these, see https://github.com/cobalt-org/liquid-rust/blob/master/src/lexer.rs#L21

PR #71, opened 2016-11-20
Add `escape_once' filter

vvv:


vvv:
```
error: unused import: `value::Value::*;`
   --> src\filters.rs:532:9
    |
532 |     use value::Value::*;
    |         ^^^^^^^^^^^^^^^^
    |
note: lint level defined here
   --> src\lib.rs:30:9
    |
30  | #![deny(warnings)]
    |         ^^^^^^^^
error: aborting due to previous error
error: Could not compile `liquid`.
```
```
error[E0432]: unresolved import `syntax::parse::token::InternedString`
 --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/clippy_lints-0.0.100/src/utils/internal_lints.rs:5:5
  |
5 | use syntax::parse::token::InternedString;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `InternedString` in `parse::token`
error[E0432]: unresolved import `syntax::parse::token::InternedString`
 --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/clippy_lints-0.0.100/src/copies.rs:6:5
  |
6 | use syntax::parse::token::InternedString;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `InternedString` in `parse::token`
```
@johannhof Could there be anything wrong with nightly rustc?

johannhof:
@vvv you can replace `use value::Value::*;` with `use value::Value::{Num, Str, Object, Array};`. That should work better.

Sorry for the delay!

src/filters.rs
```diff
@@ -390,29 +390,50 @@ pub fn modulo(input: &Value, args: &[Value]) -> FilterResult {
...
                         result.push_str(&s[last..i]);
+                        last = i + 1;
                         let escaped = match c as char {
                             '<' => "&lt;",
                             '>' => "&gt;",
-                            '&' => "&amp;",
                             '\'' => "&#39;",
                             '"' => "&quot;",
+                            '&' => {
+                                assert_eq!(skip, 0);
```
johannhof: I don't think we should do these kinds of assertions, this would only give API consumers runtime panics which is not desirable

src/filters.rs
```diff
@@ -390,29 +390,50 @@ pub fn modulo(input: &Value, args: &[Value]) -> FilterResult {
...
+                                if once_p {
+                                    skip = nr_escaped(&s[last..]);
```
johannhof: I don't get what's the deal with the `skip` variable, can't you simple have `nr_escaped` return a boolean instead and rename it `is_escaped` then? :)

vvv: Integer value specifies the number of `for` loop iterations to skip. Boolean variable would not provide such information.

src/filters.rs
vvv: The assert is harmless here. We won't reach this far if `skip` is > 0.
```
            let mut skip = 0;
            for (i, c) in s.chars().enumerate() {
                if skip > 0 {
                    skip -= 1;
                    continue;
                }
                match c as char {
[...]
```
The assert expresses our confidence in the actual value of `skip`. This value cannot be anything but 0, as explained above.

That said, if you still want me to delete `assert_eq!()`, I will. Please let me know if I should.

vvv: Sure. :)

johannhof:
Great, thanks!

PR #74, opened 2017-02-18
New filters

djwf:
I'm putting all the filters I create from the list in issue #11 here, as they are part of the standard, and the filter in pull request #73 is not.

Let me know if I should merge these two or split all filters into their own pull requests.  Also, note that I'll use this to add more filters, and will leave a note when I've finished adding filters.  If you'd prefer I not work that way, let me know and I'll change it up.

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -824,6 +824,22 @@ mod tests {
     }
 
     #[test]
+    fn unit_lstrip_non_string() {
+        let input = &Num(0f32);
+        let args = &[];
+        let desired_result = FilterError::InvalidType("Str expected".to_string());
+        assert_eq!(failed!(lstrip, input, args), desired_result);
+    }
+
+    #[test]
+    fn unit_lstrip_one_argument() {
+        let input = &tos!(" 	 \n \r test");
+        let args = &[Num(0f32)];
+        let desired_result = FilterError::InvalidArgumentCount("expected 0, 1 given".to_owned());
+        assert_eq!(failed!(lstrip, input, args), desired_result);
+    }
+
+    #[test]
     fn unit_lstrip_shopify_liquid() {
         // One test from https://shopify.github.io/liquid/filters/lstrip/
         let input = &tos!("          So much room for activities!          ");
@@ -846,22 +862,6 @@ mod tests {
         let args = &[];
         let desired_result = tos!("test 	 \n \r ");
         assert_eq!(unit!(lstrip, input, args), desired_result);
-    }
-
-    #[test]
-    fn unit_lstrip_non_string() {
-        let input = &Num(0f32);
-        let args = &[];
-        let desired_result = FilterError::InvalidType("Str expected".to_string());
-        assert_eq!(failed!(lstrip, input, args), desired_result);
-    }
-
-    #[test]
-    fn unit_lstrip_one_argument() {
-        let input = &tos!(" 	 \n \r test");
-        let args = &[Num(0f32)];
-        let desired_result = FilterError::InvalidArgumentCount("expected 0, 1 given".to_owned());
-        assert_eq!(failed!(lstrip, input, args), desired_result);
     }
 
     #[test]
@@ -1037,15 +1037,6 @@ mod tests {
     }
 
     #[test]
-    fn unit_rstrip_shopify_liquid() {
-        // One test from https://shopify.github.io/liquid/filters/rstrip/
-        let input = &tos!("          So much room for activities!          ");
-        let args = &[];
-        let desired_result = tos!("          So much room for activities!");
-        assert_eq!(unit!(rstrip, input, args), desired_result);
-    }
-
-    #[test]
     fn unit_rstrip_non_string() {
         let input = &Num(0f32);
         let args = &[];
@@ -1059,6 +1050,15 @@ mod tests {
         let args = &[Num(0f32)];
         let desired_result = FilterError::InvalidArgumentCount("expected 0, 1 given".to_owned());
         assert_eq!(failed!(rstrip, input, args), desired_result);
+    }
+
+    #[test]
+    fn unit_rstrip_shopify_liquid() {
+        // One test from https://shopify.github.io/liquid/filters/rstrip/
+        let input = &tos!("          So much room for activities!          ");
+        let args = &[];
+        let desired_result = tos!("          So much room for activities!");
+        assert_eq!(unit!(rstrip, input, args), desired_result);
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


djwf:
Whoops!  My apologies: I need to rebase this off of master, rather than off of my other pull request.  I'll do that after I get some sleep.

djwf:
@johannhof: I am confused as to which tests would go in the `src/filters.rs` file, and which would go in `tests/filters.rs`.  Can you give me some guidance on that?

djwf:
@johannhof: I'm working on `truncate`, and think liquid-rust would be well served by using proper unicode parsing of strings, thus grapheme clusters rather than characters.  May I add a dependency to an [external crate for unicode segmentation](https://kbknapp.github.io/clap-rs/unicode_segmentation/index.html) to `Cargo.toml/Cargo.lock`?  Or would you rather that happen in a separate pull request with only the filter that requires that change?

johannhof:
@djwf Thanks so much for your work on this. Just wanted to let you know it's not been forgotten, I'm simply traveling between continents for work. I'll get to reviewing this today or tomorrow. :)

djwf:
@johannhof No worries at all: I'll just keep plugging away at them.  Note that I'm using the list of filters at the [liquid markup web site](https://shopify.github.io/liquid/) as the authoritative list, as I noticed some of those missing from your list in #11 .

johannhof:
Ok, looking at this now. The re-ordering makes it a bit hard to see what exactly changed, but I'll manage. As to your questions:

- `src/filters.rs` is probably the best place to put tests. `tests/filters.rs` is intended to test cases where you want the code to go through the whole lexer, parser, etc. before running the filter (although it includes "normal" tests because I was sloppy with review)
- grapheme clusters: without having read the code yet, I agree that this is a good idea in general. I'm not sure how the Ruby version does it (as we'd ideally want to produce the same output at the end of they day), but I guess that shouldn't stop us from doing it the right way. I wonder if there are places in the lexer that could benefit from this, too. But that's irrelevant to this already pretty big PR.

Cheers

src/filters.rs
```diff
@@ -271,47 +277,191 @@ pub fn last(input: &Value, _args: &[Value]) -> FilterResult {
...
+        _ => Err(InvalidType("String expected".to_owned())),
+    }
+}
+
+/// Reverses the order of the items in an array. `reverse` cannot `reverse` a string.
+pub fn reverse(input: &Value, args: &[Value]) -> FilterResult {
+    try!(check_args_len(args, 0));
+    match *input {
+        Value::Array(ref array) => {
+            let mut reversed = array.clone();
```
johannhof: Note to self: Look if we can reduce allocation in filters (by modifying `output.rs`) before 1.0

src/filters.rs
```diff
@@ -497,6 +567,81 @@ pub fn strip_html(input: &Value, _args: &[Value]) -> FilterResult {
...
+
+    let num_chars = match args.first() {
+        Some(&Num(x)) if x > 0f32 => x as usize,
+        _ => return Err(InvalidArgument(0, "Positive number expected".to_string())),
+    };
+
+    let ellipsis = "...".to_string();
+    let append = match args.get(1) {
+        Some(&Str(ref x)) => x,
+        _ => &ellipsis,
```
johannhof: You can replace this with just `"..."`

johannhof: Ah damn I just noticed you took that from `truncatewords`. Can you please fix this up in `truncatewords`, too? Thank you!

src/filters.rs
```diff
@@ -670,6 +670,23 @@ pub fn truncatewords(input: &Value, args: &[Value]) -> FilterResult {
     }
 }
 
+/// Removes any duplicate elements in an array.
```
djwf: Got it, done.

src/filters.rs
```diff
@@ -271,47 +277,191 @@ pub fn last(input: &Value, _args: &[Value]) -> FilterResult {
...
+            Ok(Str(x.replace(arg1, arg2)))
+        }
+        _ => Err(InvalidType("String expected".to_owned())),
+    }
+}
+
+pub fn replace_first(input: &Value, args: &[Value]) -> FilterResult {
+    if args.len() != 2 {
+        return Err(InvalidArgumentCount(format!("expected 2, {} given", args.len())));
+    }
```
djwf: Got it, done.

src/filters.rs
djwf: Yep: fixed both truncate and truncatewords.

src/template.rs
```diff
@@ -1,10 +1,44 @@
 use Renderable;
 use context::Context;
-use filters::{size, upcase, downcase, capitalize, minus, plus, times, divided_by, ceil, floor,
-              round, prepend, append, first, last, pluralize, replace_first, replace, date, sort,
-              slice, modulo, escape, escape_once, remove_first, remove, strip_html, truncatewords};
```
djwf: That was just because it makes it match the list of filters in the `impl Renderable ...` section, and it made it easier to see where/what was added.  I don't have strong feelings about it either way, so I'll put it back to a grouped version (ordered alphabetically).

djwf:
Thanks for the answers and the review: note that the most recent nightly of `clippy` is complaining about the five uses of `lazy_static!` (four in lexer.rs, one in filters.rs) so (on my end, at least) the `rustup run nightly cargo clippy` fails.

Also, my rustfmt (`0.7.1 ()`) formats the grouped filter use statement as follows:

    use filters::{abs, append, capitalize, ceil, date, divided_by, downcase, escape, escape_once, first,
                  floor, join, last, lstrip, minus, modulo, newline_to_br, pluralize, plus, prepend,
                  remove, remove_first, replace, replace_first, reverse, round, rstrip, size, slice,
                  sort, split, strip, strip_html, strip_newlines, times, truncate, truncatewords, uniq,
                  upcase};

The rustfmt for the Travis CI builds, however, formats the grouped filter use statement differently:

    use filters::{abs, append, capitalize, ceil, date, divided_by, downcase, escape, escape_once,
                  first, floor, join, last, lstrip, minus, modulo, newline_to_br, pluralize, plus,
                  prepend, remove, remove_first, replace, replace_first, reverse, round, rstrip, size,
                  slice, sort, split, strip, strip_html, strip_newlines, times, truncate,
                  truncatewords, uniq, upcase};

No big deal, but I figured you'd want to know.

johannhof:
> note that the most recent nightly of clippy is complaining about the five uses of lazy_static! (four in lexer.rs, one in filters.rs) so (on my end, at least) the rustup run nightly cargo clippy fails.

Seems like there's an issue around that recently added rule: https://github.com/Manishearth/rust-clippy/issues/1580

You can ignore it by adding `#![cfg_attr(feature="cargo-clippy", allow(zero_ptr))]` to `lib.rs`.

> The rustfmt for the Travis CI builds, however, formats the grouped filter use statement differently:

Feel free to update the version to `v0.7.1-20170120` here: https://github.com/cobalt-org/liquid-rust/blob/master/.travis.yml#L10

johannhof:
> You can ignore it by adding #![cfg_attr(feature="cargo-clippy", allow(zero_ptr))] to lib.rs.

Nevermind, I just merged #80 so you'll just have to rebase :)

johannhof:
Looks great!

PR #76, opened 2017-02-25
add default filter

fiji-flo:
More to come. It's not covered in #74 and missing in #11 (see my comment there).

Mimics behavior from the [shopify implementation](https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb#L356)


fiji-flo:
resolved merge conflicts

johannhof:
Cool, thanks!

PR #77, opened 2017-02-26
whitespace control

### LINKED ISSUE #67: Support whitespace control
Not sure when they added that but we should support this: https://shopify.github.io/liquid/basics/whitespace/

fiji-flo:
resolves #67

Unit tests added. Shouldn't affect anything outside the lexer. Please confirm / check this.

### DIFF
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -18,7 +18,11 @@ pub enum Element {
 }
 
 lazy_static! {
-    static ref MARKUP: Regex = Regex::new("\\{%.*?%\\}|\\{\\{.*?\\}\\}").unwrap();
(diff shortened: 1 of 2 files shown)


fiji-flo:
There seems to be an issue with clippy and the latest nightly (all [clippy prs](https://github.com/rust-lang/rust/pull/40026#issuecomment-282531359) are failing, too).

I couldn't reproduce it locally sine [rustup for nightly is broken](https://github.com/rust-lang/rust/issues/40106), too.

johannhof:
Sorry for the delay. This is quite clever! Great job. If the tests pass the lexer should be fine, otherwise we need to fix it and add more tests ;)

PR #78, opened 2017-02-26
Only capitalize the first character

dzamlo:
According to https://shopify.github.io/liquid/filters/capitalize/, the `capitalize` filter should only capitalize the first character.


johannhof:
Looks good, thank you!

PR #81, opened 2017-02-28
Fix for cobalt.rs categories

uwearzt:
For use with https://github.com/cobalt-org/cobalt.rs/pull/196


epage:
Since the cobalt issue is closed, should this be closed or is there some value this still provides?  If so, what functionality are you trying to accomplish with this?  It seems like its trying to look up the key in an object inside of an array but I'm not sure why that is needed.

uwearzt:
I will work on the categories fix, just renamed the branch and starting over :)

PR #82, opened 2017-03-31
fix(appveyor): Correct the path to curl

epage:


epage:
*sigh* Travis died again because of clippy, merging anyways.

PR #90, opened 2017-04-30
File system trait and template locations

gmalette:
Hello! 

This is my first Rust contribution, so I expect it to be less-than-perfect, please let me know how I can improve it!

---

Currently `LiquidOptions.file_system` is implemented as a `PathBuf`, reading directly for the disk. This leaves little options for where to store the files. I'd like to have my files in abstract repositories, either in the database, S3, etc. As a result, I changed `file_system` to be a trait, boxed in `LiquidOptions`. 

Then I realized my custom blocks needed a way to specify to the file system where the templates were stored. This required little change, but I still thought it would be a good time to introduce a `TemplateName` type. I wondered if I should change `include_tag` to send `include://#{path}` as the `TemplateName`, like I would do in custom blocks, but I ended up not doing it. Thoughts on this?

Finally, I wondered if `file_system` was really a `file_system` given the previous changes. I renamed it to `template_repository`. I realize that this is diverging from Shopify/liquid, so I don't have a particularly strong opinion on this one.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,9 +126,8 @@ pub trait Renderable: Send + Sync {
     fn render(&self, context: &mut Context) -> Result<Option<String>>;
 }
 
-pub type TemplateName = String;
 pub trait TemplateRepository {
-    fn read_template(&self, path: &TemplateName) -> Result<String>;
+    fn read_template(&self, path: &str) -> Result<String>;
 }
 
 /// `TemplateRepository` to load files relative to the root
@@ -143,7 +142,7 @@ impl LocalTemplateRepository {
 }
 
 impl TemplateRepository for LocalTemplateRepository {
-    fn read_template(&self, relative_path: &TemplateName) -> Result<String> {
+    fn read_template(&self, relative_path: &str) -> Result<String> {
         let path = self.root.clone().join(relative_path);
 
         if !path.exists() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,15 +126,48 @@ pub trait Renderable: Send + Sync {
     fn render(&self, context: &mut Context) -> Result<Option<String>>;
 }
 
+pub trait FileSystem {
+    fn read_template_file(&self, path: &Path) -> Result<String>;
+}
+
+/// FileSystem to load files relative to the root
+pub struct LocalFileSystem {
+    root: PathBuf,
+}
+
+impl FileSystem for LocalFileSystem {
+    fn read_template_file(&self, relative_path: &Path) -> Result<String> {
+        let path = self.root.clone().join(relative_path);
+
+        if !path.exists() {
+            return Err(Error::from(&*format!("{:?} does not exist", path)));
+        }
+        let mut file = try!(File::open(path));
+
+        let mut content = String::new();
+        file.read_to_string(&mut content)?;
+        Ok(content)
+    }
+}
+
 /// Options that `liquid::parse` takes
-#[derive(Default)]
 pub struct LiquidOptions {
     /// Holds all custom block-size tags
     pub blocks: HashMap<String, Box<Block>>,
     /// Holds all custom tags
     pub tags: HashMap<String, Box<Tag>>,
     /// The path to which paths in include tags should be relative to
-    pub file_system: Option<PathBuf>,
+    pub file_system: Box<FileSystem>,
+}
+
+impl Default for LiquidOptions {
+    fn default() -> LiquidOptions {
+        LiquidOptions {
+            blocks: Default::default(),
+            tags: Default::default(),
+            file_system: Box::new(LocalFileSystem { root: PathBuf::new() }),
+        }
+    }
 }
 
 impl LiquidOptions {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,8 +126,9 @@ pub trait Renderable: Send + Sync {
     fn render(&self, context: &mut Context) -> Result<Option<String>>;
 }
 
+pub type TemplateLocation = Path;
 pub trait FileSystem {
-    fn read_template_file(&self, path: &Path) -> Result<String>;
+    fn read_template_file(&self, path: &TemplateLocation) -> Result<String>;
 }
 
 /// FileSystem to load files relative to the root
@@ -136,7 +137,7 @@ pub struct LocalFileSystem {
 }
 
 impl FileSystem for LocalFileSystem {
-    fn read_template_file(&self, relative_path: &Path) -> Result<String> {
+    fn read_template_file(&self, relative_path: &TemplateLocation) -> Result<String> {
         let path = self.root.clone().join(relative_path);
 
         if !path.exists() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,16 +127,16 @@ pub trait Renderable: Send + Sync {
 }
 
 pub type TemplateLocation = Path;
-pub trait FileSystem {
+pub trait FileRepository {
     fn read_template_file(&self, path: &TemplateLocation) -> Result<String>;
 }
 
 /// FileSystem to load files relative to the root
-pub struct LocalFileSystem {
+pub struct LocalFileRepository {
     root: PathBuf,
 }
 
-impl FileSystem for LocalFileSystem {
+impl FileRepository for LocalFileRepository {
     fn read_template_file(&self, relative_path: &TemplateLocation) -> Result<String> {
         let path = self.root.clone().join(relative_path);
 
@@ -158,7 +158,7 @@ pub struct LiquidOptions {
     /// Holds all custom tags
     pub tags: HashMap<String, Box<Tag>>,
     /// The path to which paths in include tags should be relative to
-    pub file_system: Box<FileSystem>,
+    pub file_repository: Box<FileRepository>,
 }
 
 impl Default for LiquidOptions {
@@ -166,7 +166,7 @@ impl Default for LiquidOptions {
         LiquidOptions {
             blocks: Default::default(),
             tags: Default::default(),
-            file_system: Box::new(LocalFileSystem { root: PathBuf::new() }),
+            file_repository: Box::new(LocalFileRepository { root: PathBuf::new() }),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,18 +126,24 @@ pub trait Renderable: Send + Sync {
     fn render(&self, context: &mut Context) -> Result<Option<String>>;
 }
 
-pub type TemplateLocation = Path;
-pub trait FileRepository {
-    fn read_template_file(&self, path: &TemplateLocation) -> Result<String>;
+pub type TemplateName = String;
+pub trait TemplateRepository {
+    fn read_template(&self, path: &TemplateName) -> Result<String>;
 }
 
-/// FileSystem to load files relative to the root
-pub struct LocalFileRepository {
+/// `TemplateRepository` to load files relative to the root
+pub struct LocalTemplateRepository {
     root: PathBuf,
 }
 
-impl FileRepository for LocalFileRepository {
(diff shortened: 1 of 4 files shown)


epage:
*sigh* The CI is being a pain
- Travis has a broken Clippy
- Appveyor has the Nul problem.

gmalette:
Any idea what I can do to get CI passing? You seem to have a 💚 on master.

epage:
Now that we have enough coverage, I'm fine ignoring the last CI failure.

While I like the idea of your change, I've not had a chance to dig through the implementation.  Sorry about that.  Hopefully tonight.

tests/custom_blocks.rs
```diff
@@ -36,11 +36,7 @@ fn run() {
         Ok(Box::new(Multiply { numbers: numbers }))
     }
 
-    let mut options = LiquidOptions {
-        blocks: Default::default(),
-        tags: Default::default(),
-        file_system: Default::default(),
-    };
+    let mut options: LiquidOptions = Default::default();
```
epage: Thanks!

src/lib.rs
```diff
@@ -126,15 +126,49 @@ pub trait Renderable: Send + Sync {
     fn render(&self, context: &mut Context) -> Result<Option<String>>;
 }
 
+pub type TemplateLocation = Path;
```
epage: If the intention is to abstract away where the snippets are coming from, I feel like keeping it as a `Str` rather than `Path` is ideal.

Also `Location` in the name doesn't quite sit right with me.  What do you think of something like `Name` or `Path`

src/lib.rs
```diff
@@ -126,15 +126,49 @@ pub trait Renderable: Send + Sync {
...
+pub trait FileRepository {
```
epage: I feel like calling it a `FileRepository` is leaking the implementation details.  Maybe `TemplateRepository`?

Now for some more specific parts of the name

`Template`: I think this will be fine.  Upon reviewing this, I looked up what liquid refers to what `include` pulls in.  They don't formally name their parameters for the `include` tag that I saw but referred to it as a snippet.  Their source code refers to it as a template.  Either way, I have bigger ambitions for this.  I think it'd be useful for caching at which point it will be reused both for parsing top level files and for parsing `include`d files.  At that point, `Template` works.

`Repository`: I think this will be fine.  My gut tells me that there are some common terms for this type of thing but I can't think of them atm.  Maybe `Repository` is one.  Either way, if we don't have another idea for a name, I think this `Repository` will be fine.

src/lib.rs
```diff
@@ -126,15 +126,49 @@ pub trait Renderable: Send + Sync {
...
+    fn read_template_file(&self, path: &TemplateLocation) -> Result<String>;
```
epage: I think there is a valuable opportunity for caching here.

I'm just trying o think of what should be cached.  Could we make it so the `Template` gets cached?  Do we need to instead somehow expose `Vec<Element>` (ie token stream) and cache that?

Any thoughts?  If you don't mind the break, this is a bigger thing that could be deferred to later.

src/lib.rs
```diff
@@ -126,15 +126,49 @@ pub trait Renderable: Send + Sync {
...
 /// Options that `liquid::parse` takes
-#[derive(Default)]
 pub struct LiquidOptions {
     /// Holds all custom block-size tags
     pub blocks: HashMap<String, Box<Block>>,
     /// Holds all custom tags
     pub tags: HashMap<String, Box<Tag>>,
     /// The path to which paths in include tags should be relative to
-    pub file_system: Option<PathBuf>,
+    pub file_repository: Box<FileRepository>,
```
epage: Box or ref?

I think box is fine.  A small part of me wishes this was instead a ref but the lifetime would then have to be added through the entire code base and might get nasty.  Maybe something we can re-evaluate for later.

epage:
Thanks for working on this and again, I'm sorry for the delay in reviewing it!

src/lib.rs
epage: `Repository`: `Factory` is the word I was looking for.  Either works for me.

src/lib.rs
epage: Side note: languages like Rust are awesome but they always make me want more.  In cases like yours, transferring ownership is the optimal solution. In caching cases, borrows are ideal.  Its hard to satisfy both.  I guess that is what `Cow` is for.  Unsure if we should use `Cow` or not.

gmalette:
Im currently in the process of moving, I'll be back to this on Sunday.
Thanks for the very thorough review !

src/lib.rs
gmalette: Wouldn't the ref have to be parameterized in the `LiquidOption`'s type? I had started this way, but then it occurred to me that it's backwards that the world should know about the options, so I elected the box.

src/lib.rs
gmalette: I don't think the FS layer should return a token stream, but we could have a separate plumping in between the two.

What do you think of the following:

`LiquidRendering -> TemplateFactory -> TemplateRepository`

If you like this structure, then the `LiquidOptions` would take a `TemplateFactory`, that would return something higher-level than a string; probably a template.

src/lib.rs
gmalette: Shopify/liquid historically called included templates `snippet`s. More recently, they've added custom blocks/tags that load templates from other folders, such as [`section`](https://help.shopify.com/themes/liquid/tags/theme-tags#section).

After this PR, I'd really like liquid-rust to support non-`include` that also load templates.

src/lib.rs
gmalette: I'm not sure what caching you're referring to in this case. 

If we're trying to avoid disk access, one could compose a `TemplateFactory` that has one FS layer and one in-memory layer for caching. 

If you're trying to avoid copying memory, we could be lending refs

src/lib.rs
epage: It seems there is a difference between public Liquid and Liquid for Shopify's customers.  I'm trying to better understand it.  Not sure how we should express the difference in our liquid implementation.

These different type of template load systems seem to use different directories (from my quick glance at the `section` documentation).  I wonder if your new trait should take two parameters, a type and a name.  Not sure if the type should be a string or an enum.

src/lib.rs
epage: Yeah, for now, I feel a `Box` is better.  As I'm learning the code, I'm considering how I want to refactor it.  Maybe after the refactoring, what should be done will become more clear.

My idea is to turn LiquidOptions into a Parser (a Template Builder).  I plan to move filters into the Parser and exposing clients to set their own filters.  Not sure what else will fall out of these changes.

src/lib.rs
epage: My goal with caching is to reduce disk access + parse time when batch processing templates like in cobalt.

Could you elaborate on your idea for `LiquidRendering -> TemplateFactory -> TemplateRepository`?  What is the role of each, how do they compose, etc?

gmalette: Sure! I think there's one missing concept right now in liquid-rust. Let's ignore that for now and see how `TemplateFactory` and `TemplateRepository` are used together. 

```rust
// This may not be idiomatic Rust, sorry for that!

// The TemplateRepository knows how to find the template content 
// either on the disk, in the database, on S3, whatever.
trait TemplateRepository {
  fn fetch(template_name: TemplateName) -> Option<Result<String>>;
}

// This is the current implementation in liquid-rust
// It uses a basic default path, and naively appends
// the template name.
struct LocalTemplateRepository {
  root: PathBuf,
}

impl TemplateRepository for LocalTemplateRepository {
  fn fetch(template_name: TemplateName) -> Option<Result<String>> {
    // read from file system
  }
}

// We could also provide a default caching repository.
// If it's cache doesn't contain an entry for the template,
// it would use it's backend to fetch it, then grab ownership
// of the template, and lend it when a template needs to be 
// rendered.
// I'm not very versed in lifetimes (haven't had to deal with them
// yet, sorry again :)
struct<'a> TemplateRepositoryCache<'a> {
  cache: MapOfSomeSort<TemplateName, &'a Template>,
  backend: Box<TemplateRepository>,
}

impl TemplateRepository for TemplateRepositoryCache {
  cache.entry(template).or_insert_with(|| self.backend.generate(...))
}

// You could also have a TemplateRepository that uses multiple
// other repositories, falling back to the others if the template
// is not found.
struct CascadingTemplateRepository {
  backends: Vec<Box<TemplateRepository>>,
}

// The TemplateFactory is the equivalent of today's `parse` function. 
// It takes a repository to find the template contents, and returns the 
// templates themselves. 
trait TemplateFactory {
  pub fn generate(&self, template_name: TemplateName, options: LiquidOptions) -> Option<Result<&Template>>
}

// The only difference with the current way is that it uses a repository
// to locate the template contents
struct DefaultTemplateFactory {
  template_repository: Box<TemplateRepository>,
};

impl TemplateFactory for DefaultTemplateFactory {
  pub fn generate(...) -> Options<Result<&Template>> {
    let template_content = self.template_repository.fetch(template_name);
    parse(template_content, options)
  } 
}

// Same as for the repository, the factory can have a naive cache.
struct TemplateFactoryCache {
  cache: MapOfSomeSort,
  backend: Box<TemplateFactory>,
}

impl TemplateFactory for TemplateFactoryCache {
  pub fn generate(...) -> Option<Result<&Template>> {
    cache.entry(template).or_insert_with(|| self.backend.generate(...))
  }
}

let repository = S3TemplateRepository::new(my_s3_bucket);
let factory = DefaultTemplateFactory::new(repository);
let options = LiquidOptions::new(factory, ...);
let template = factory.generate(template_name, options);
```

The problem I see with this isn't a dealbreaker, but it bugs me that the factory needs to know about the options, and the options about the factory. That's where the `Renderer` comes in.

```rust
let liquid_renderer = LiquidRenderer::new(factory, options);
let output = liquid_renderer.render(template_name, &mut context);
```

epage: You bring up some good points.  This is going to be a bigger decision that we should probably plan out in a separate RFC issue.  How about we keep to just returning a string for now?

gmalette:
I went back to something close to the original PR, renaming `FileRepository` to `TemplateRepository`. I cannot, for the life of me, get the `cargo fmt` step of travis to pass. I don't know what I doing wrong, but locally I get totally different results for `cargo fmt` than what CI shows. Any hint?

epage:
> I cannot, for the life of me, get the cargo fmt step of travis to pass. I don't know what I doing wrong, but locally I get totally different results for cargo fmt than what CI shows. Any hint?

Originally our CI was stuck on a specific version of rustfmt.  I'm playing with us always using the latest now.  I'm updating it now to at least report the version of rustfmt to use.

epage:
Also, I probably won't get to reviewing your latest changes until tonight or tomorrow night.  I'm sorry for the delay.

gmalette:
> I'm sorry for the delay.

No worries! You're already quicker than I expected!

epage:
For the CI, I've locked rustfmt to `0.8.3`.  If you could use that version to make everything happy (there has got to be a better way).

gmalette:
Sorry the PR was closed because of a bad rebase (or rather I accidentally used `reset` instead of `rebase`)

src/lib.rs
```diff
@@ -126,15 +126,55 @@ pub trait Renderable: Send + Sync {
     fn render(&self, context: &mut Context) -> Result<Option<String>>;
 }
 
+pub type TemplateName = String;
+pub trait TemplateRepository {
+    fn read_template(&self, path: &TemplateName) -> Result<String>;
```
epage: Drat, clippy isn't happy about TemplateName references.

Could we change TemplateName to be a typdef for str?  If not, I'm fine getting rid of the typedef for now.

src/lib.rs
epage: btw for historical reference, the RFC issue is #95

PR #97, opened 2017-05-14
refactor(Parse): Rename VarOrVal to Argument

gmalette:
I was trying to implement named arguments for filters, and I had trouble understanding what `VarOrVal` meant. I understood that it could contain either a variable or a value, but understanding the union of those types was hard.

It occurred to me that holding these values is just a coincidence; what `VarOrVal` really represents, is the arguments passed into filters [as used in FilterPrototype](https://github.com/cobalt-org/liquid-rust/compare/master...gmalette:rename-varorval-to-argument?expand=1#diff-ac5cd937cc00cb0544b06a1b6ce1ce5eR10). It's the "unknown" value of an argument between the time the code gets parsed until it gets evaluated.

Thoughts about renaming it `FilterArgument` ?

Had the grammar been structured different, I would've suggested something more along the lines of

```rust
enum FilterArgument {
  Literal(Value)
  Identifier(Variable)
}
```

Speaking of grammar, I'm curious: do you know why the tokenizer and parser were written by hand, rather than using a generator?

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -7,17 +7,17 @@ use error::{Error, Result};
 #[derive(Debug, PartialEq)]
 pub struct FilterPrototype {
     name: String,
-    arguments: Vec<VarOrVal>,
+    arguments: Vec<Argument>,
 }
 
 #[derive(Debug, PartialEq)]
-pub enum VarOrVal {
+pub enum Argument {
     Var(Variable),
     Val(Value),
 }
 
 impl FilterPrototype {
-    pub fn new(name: &str, arguments: Vec<VarOrVal>) -> FilterPrototype {
+    pub fn new(name: &str, arguments: Vec<Argument>) -> FilterPrototype {
         FilterPrototype {
             name: name.to_owned(),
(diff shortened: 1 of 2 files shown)


gmalette:
I've been thinking about this a bit more... do we want tags to use these for their own parsing?

```rust
pub fn decrement_tag(tag_name: &str, args: &[Token], ...) {
  let mut args: Vec<Argument> = parse_arg_list(args.iter())
  ...
}
```

In that case, we should probably call it `enum Argument` instead of `FilterArgument``

epage:
> Thoughts about renaming it FilterArgument ?
> In that case, we should probably call it enum Argument instead of FilterArgument

I think that is reasonable.

> Speaking of grammar, I'm curious: do you know why the tokenizer and parser were written by hand, rather than using a generator?

My best guess is that it made writing liquid-rust faster because it involved just porting the Ruby regexes to Rust regexes.  I'd love to see it switched to a proper grammar.

gmalette:
Done

PR #116, opened 2017-05-27
feature(Filter): Add keyword arguments

### LINKED ISSUE #92: Support for named arguments in filters
Depends on #301 

Shopify/liquid supports named arguments for filters. 

[Example](https://help.shopify.com/themes/liquid/filters/additional-filters#default_pagination)

```liquid
{{ paginate | default_pagination: next: 'Older', previous: 'Newer' }}
```

Is this something liquid-rust aims to support? If yes, I'd be happy to make a PR for this; I already have part of a PR.

gmalette:
Fixed #92

This PR adds named arguments to filters, much like Shopify/liquid does; using a 3rd argument to `FilterPrototype`.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -8,6 +8,7 @@ use error::{Error, Result};
 pub struct FilterPrototype {
     name: String,
     arguments: Vec<Argument>,
+    named_arguments: Vec<(String, Argument)>,
 }
 
 #[derive(Debug, PartialEq)]
@@ -17,10 +18,14 @@ pub enum Argument {
 }
 
 impl FilterPrototype {
-    pub fn new(name: &str, arguments: Vec<Argument>) -> FilterPrototype {
+    pub fn new(name: &str,
+               arguments: Vec<Argument>,
+               named_arguments: Vec<(String, Argument)>)
+               -> FilterPrototype {
         FilterPrototype {
             name: name.to_owned(),
             arguments: arguments,
+            named_arguments: named_arguments,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/output.rs
```diff
@@ -17,10 +18,14 @@ pub enum Argument {
 }
 
 impl FilterPrototype {
-    pub fn new(name: &str, arguments: Vec<Argument>) -> FilterPrototype {
+    pub fn new(name: &str,
+               arguments: Vec<Argument>,
+               named_arguments: Vec<(String, Argument)>)
```
gmalette: I was hesitant between `Vec<(String, Argument)>` and `Map<(String, Argument)>`.
Suggestions?

src/parser.rs
```diff
@@ -55,65 +55,103 @@ fn parse_expression(tokens: &[Token], options: &LiquidOptions) -> Result<Box<Ren
...
+        return Ok(Output::new(entry, vec![]));
+    }
+
+    expect(&mut iter, &Pipe)?;
+
     let mut filters = vec![];
+
+    for filter_tokens in tokens.split(|token| token == &Pipe).skip(1) {
+        filters.push(try!(parse_filter(filter_tokens)));
+    }
```
gmalette: I had a lot of difficulty with this because at first I thought named arguments had to be at the end of the call (since the filter gets them as the 3rd arg). However, further investigation revealed that named args can be followed by positional args, which meant that this refactor wasn't needed. 

Splitting the filter chain for parsing means it was easier for me to grok, however it produces less-than-ideal error messages when an unexpected token is present at the end of the filter's token slice, as shown in [this test](https://github.com/cobalt-org/liquid-rust/pull/116/files#diff-2c09afcdc3c420ab0678ba9b5e83959cR363)

PR #117, opened 2017-05-27
feature(Filter): Add named arguments

### LINKED ISSUE #92: Support for named arguments in filters

gmalette:
Fixed #92

This PR adds named arguments to filters, much like Shopify/liquid does; using a 3rd argument to `FilterPrototype`.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -8,6 +8,7 @@ use error::{Error, Result};
 pub struct FilterPrototype {
     name: String,
     arguments: Vec<Argument>,
+    named_arguments: Vec<(String, Argument)>,
 }
 
 #[derive(Debug, PartialEq)]
@@ -17,10 +18,14 @@ pub enum Argument {
 }
 
 impl FilterPrototype {
-    pub fn new(name: &str, arguments: Vec<Argument>) -> FilterPrototype {
+    pub fn new(name: &str,
+               arguments: Vec<Argument>,
+               named_arguments: Vec<(String, Argument)>)
+               -> FilterPrototype {
         FilterPrototype {
             name: name.to_owned(),
             arguments: arguments,
+            named_arguments: named_arguments,
         }
     }
 }
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -184,8 +184,9 @@ fn test_whitespace_control() {
 fn test_split_atom() {
     assert_eq!(split_atom("truc | arg:val"),
                vec!["truc", " ", "", "|", "", " ", "arg", ":", "val"]);
-    assert_eq!(split_atom("truc | filter:arg1,arg2"),
-               vec!["truc", " ", "", "|", "", " ", "filter", ":", "arg1", ",", "arg2"]);
+    assert_eq!(split_atom("truc | filter:arg1,arg2,a:arg3"),
+               vec!["truc", " ", "", "|", "", " ", "filter", ":", "arg1", ",", "arg2", ",", "a",
+                    ":", "arg3"]);
 }
 
 #[test]
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -65,7 +65,7 @@ pub fn parse_output(tokens: &[Token]) -> Result<Output> {
     };
 
     if let None = iter.peek() {
-        return Ok(Output::new(entry, vec![]))
+        return Ok(Output::new(entry, vec![]));
     }
 
     expect(&mut iter, &Pipe)?;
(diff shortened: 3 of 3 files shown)


epage:
Working on a fix for nightly's build failure.

src/output.rs
```diff
@@ -17,10 +18,14 @@ pub enum Argument {
...
 impl FilterPrototype {
-    pub fn new(name: &str, arguments: Vec<Argument>) -> FilterPrototype {
+    pub fn new(name: &str,
+               arguments: Vec<Argument>,
+               named_arguments: Vec<(String, Argument)>)
+               -> FilterPrototype {
         FilterPrototype {
             name: name.to_owned(),
             arguments: arguments,
+            named_arguments: named_arguments,
```
epage: So this change will allow parsing of `named_arguments` but we don't seem to apply them (see `Output::apply_filters`).
- What is the plan for actually applying the named arguments?
- What exactly are the semantics for named arguments that we need to support, like mixing positional and named?  How can we make this easy for filter writers?  How do ensure performance?

epage:
> Working on a fix for nightly's build failure.

Fixed, see #119

src/output.rs
gmalette: 🤦‍♂️

> What is the plan for actually applying the named arguments?

I'm not sure what you mean by "applying". I guess we'll have to change the signature of the fn for the tags, introducing a breaking change :(. 

> What exactly are the semantics for named arguments that we need to support, like mixing positional and named?

At first I started by only supporting named args at the end of the filter, but then I had the brilliant idea of checking how Shopify/liquid does it. [As you can see](https://github.com/Shopify/liquid/blob/c582b86f16b64f5df890807603f57914099f3868/lib/liquid/variable.rb#L93-L99), named args and positional args can be mixed.

> How can we make this easy for filter writers?

I've been flip-flopping on this. At first I thought of using a `HashMap`. After all, keys are always strings. Then I figured Erlang just uses lists of tuples, which for low amount of args is probably more performant. Now I don't know. For ease of writing, I think `HashMap`.

epage: >  What exactly are the semantics for named arguments that we need to support, like mixing positional and named?
> At first I started by only supporting named args at the end of the filter, but then I had the brilliant idea of checking how Shopify/liquid does it. As you can see, named args and positional args can be mixed.

Some things we need to consider
- When mixing named and positional, how does the position index change?  If I do `val0, arg1: val1, valN`, is `valN` `val1` or `val2`?
- Repeating a named argument: `arg0: val0, arg0: something`
- Repeating a positional argument as a named argument: `val0, val1, val2, arg0: something`

epage: > I'm not sure what you mean by "applying". I guess we'll have to change the signature of the fn for the tags, introducing a breaking change :(.

That is what I mean.  Currently, the named arguments aren't being passed to the filter.

epage: > How can we make this easy for filter writers?

HashMap vs list of tuples vs some kind of flat map (list of tuples with a Map API) don't really matter except for performance as long as we provide a stable API on hitting 1.0.

For how to make it easier on filter writers, one idea I just had is for a filter to do the following
```rust
let required = ["param0", "param1"];
let optional = ["param2", "param3"];
check_args_len(args, required, optional)?;

let index = Index(args, required, optional);
let offset = index[0]...;
let length = index.get(1)...;
```

The Index object implements all of the lookup rules for the developer.  Positional arguments are passed in for fast access for positional arguments but if the positional argument isn't there, the `required` and `optional` lists are indexed to find the name to look up the value in `named_arguments`.

This actually gets me thinking. 
- Clients need to rely on our helper code to get the correct semantics.  This means we need to expose it and hope they use it
- By exposing helper code, we are increasing our compatibility surface

I wonder if what we should have filters register `required` and `optional` arrays along with their `name` and `fn`.  We then have several options
- Pass the `Index` object (probably renamed to `Arguments`) to a filter
- Merge `named_arguments` into the unnamed arguments, passing one `Vec` in.  We then need to deal with holes in optional arguments caused by named arguments (`opt0: val0, opt3: val3`, no `opt1` or `opt2` defined).  We could handle this by putting all arguments into an `Option` or passing two `Vec`s to a filter, one for required and one for optional parameters.

epage:
> That being said, I've pretty much put the project on ice after semi-failing to do named args :(. I really appreciate you looking out for me, but at this point it's wiser to focus your efforts on your own needs.

I think I missed it. In what way did this semi-fail? Is there anything I can do to help?

gmalette:
I don't recall exactly, but something along: I had trouble adapting the tokenizer or parser for the named args, the `next()` didn't work very well and I would've need a double-peek iterable... or something like that.

In a previous discussion, you hinted you'd be open to using a parser generator, so I tried to write it in Nom (https://github.com/gmalette/liquid-rust/pull/2) but I couldn't generate good error messages.

Having failed both these approaches, I kinda gave up 😢

epage:
😢 indeed

Thanks for the explanation so that others who come along can know what problems they'll need to overcome to get this to work.

Goncalerta:
Closing this as named arguments was added in #319

PR #122, opened 2017-06-11
fix(CI): rustfmt dies on nightly

epage:


epage:
Ideally we'd get a binary of rustfmt but its hard to get users and CI to match

PR #130, opened 2017-09-22
Workaround for skeptic causing liquid crate to recompile every time

jpastuszek:
Skeptic causes liquid to recompile every time.
This workaround stops cargo from refreshing the build.
See: 
* https://github.com/budziq/rust-skeptic/issues/25
* https://github.com/rust-lang/cargo/issues/4213


epage:
*sigh* The error chain problem.

epage:
Also, while I know the details are in those links on your PR, could you link to them in your commit message and summarize them (skeptic tells cargo about a file most people use, forcing cargo to always rebuild in the vain hope that it will actually be code-genned this time).

PR #131, opened 2017-09-22
Workaround for skeptic causing liquid crate to recompile every time

jpastuszek:
Build script calls `skeptic` which emits information asking cargo to monitor content of `README.md.skt.md` for changes. If changes to that file are detected cargo should rebuild `liquid` crate.
In case of this project the files does not exist. This causes cargo to trigger a rebuild every time in hope that it will get created by the build script.
The workaround is to create the file that skeptic asks to monitor which stops cargo from triggering the rebuild.

For details consult this issues:

* https://github.com/budziq/rust-skeptic/issues/25
* https://github.com/rust-lang/cargo/issues/4213

To reproduce the issue run `cargo test` on project depending on 'liquid' multiple times and observe that `liquid` crate is rebuild every time. After this fix is applied `cargo test` should only compile `liquid` crate once.


jpastuszek:
Same again with more info.

PR #140, opened 2017-10-27
feat(value): Add null value

gnunicorn:
This adds the missing null type to the values of liquid, thus allowing to
parse serde_json and others containing null-values in the data directly

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -16,6 +16,7 @@ pub enum Value {
     Str(String),
     Array(Array),
     Object(Object),
+    Nil,
 }
 
 /// Type representing a Liquid array, payload of the `Value::Array` variant
@@ -132,6 +133,7 @@ impl PartialEq<Value> for Value {
             (&Value::Str(ref x), &Value::Str(ref y)) => x == y,
             (&Value::Array(ref x), &Value::Array(ref y)) => x == y,
             (&Value::Object(ref x), &Value::Object(ref y)) => x == y,
+            (&Value::Nil, &Value::Nil) => true,
 
             // encode Ruby truthiness; all values except false and nil
             // are true, and we don't have a notion of nil
@@ -164,6 +166,7 @@ impl ToString for Value {
             Value::Num(ref x) => x.to_string(),
             Value::Bool(ref x) => x.to_string(),
             Value::Str(ref x) => x.to_owned(),
+            Value::Nil => "".to_owned(),
             Value::Array(ref x) => {
                 let arr: Vec<String> = x.iter().map(|v| v.to_string()).collect();
                 arr.join(", ")
@@ -267,6 +270,11 @@ mod test {
     fn numbers_have_ruby_truthiness() {
         assert_eq!(TRUE, Value::Num(42f32));
         assert_eq!(TRUE, Value::Num(0f32));
+    }
+
+    #[test]
+    fn nil_equality() {
+        assert_eq!(Value::Nil, Value::Nil);
     }
 
     #[test]
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -849,6 +849,7 @@ pub fn default(input: &Value, args: &[Value]) -> FilterResult {
         Array(ref a) => a.is_empty(),
         Bool(b) => !b,
         Num(_) => false,
+        Value::Nil => true,
     };
 
     if use_default {
--- a/tests/value.rs
+++ b/tests/value.rs
@@ -67,6 +67,22 @@ pub fn deserialize_bool() {
 }
 
 #[test]
+pub fn serialize_nil() {
+    let actual = liquid::Value::Nil;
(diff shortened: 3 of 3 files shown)


epage:
My hope has been to avoid the [billion dollar mistake](https://en.wikipedia.org/wiki/Tony_Hoare#Apologies_and_retractions).

Is there a use case for you to have null in your data?

gnunicorn:
Well, this is for parsing data in cobalt, where any json can contain a `null` value and serde currently refuses to load the data into value then -- giving you rather obscure errors.

I don't think we are repeating that case here but are downwards compatible to those formats. We aren't repeating that problem, as all values have a way to be checked for being empty and will before interacting with it and `null`.

I see your point that this might not be something we want in liquid though and therefore should be handled inside cobalt (where I'd argue, we should be able to parse any valid JSON and `null` is a valid JSON value). If you agree on that we might want to discuss how to do that though: can we extend the serde-parsing into `Value` from cobalt? Then we could just use `false`. But I'd need some hints on how to achieve that.

epage:
What is the reason for having null in your cobalt data?  Is it because you are loading a existing data source and don't plan to touch the null part of the data?  Do you need to touch the null part and need liquid to handle it gracefully?

> I see your point that this might not be something we want in liquid though and therefore should be handled inside cobalt (where I'd argue, we should be able to parse any valid JSON and null is a valid JSON value). 

Really, the question is whether we are loading arbitrary json/yaml/toml or loading liquid values.   My initial impression is to load liquid Values unless someone has a reasonable use case that can't be supported.

epage:
> for AreWebYet I am just downloading crates.io JSON-dumps of packages and put them into the data folder. Those can and do contain null values (e.g. https://github.com/bashyHQ/arewewebyet/blob/gh-pages/_data/pkgs/afterparty.json )

Thanks for the use case.  I am fine moving forward with this.

gnunicorn:
@epage it's green.

src/value.rs
```diff
@@ -164,6 +165,7 @@ impl ToString for Value {
             Value::Num(ref x) => x.to_string(),
             Value::Bool(ref x) => x.to_string(),
             Value::Str(ref x) => x.to_owned(),
+            Value::Null => "null".to_owned(),
```
epage: Sorry, just thought about this last night.  Are we aligned with Shopify on this?

Does it matter?

src/value.rs
```diff
@@ -16,6 +16,7 @@ pub enum Value {
     Str(String),
     Array(Array),
     Object(Object),
+    Null,
```
epage: I wonder how serde decides that `None` gets serialized as this.

We should add a serde test for this
https://github.com/cobalt-org/liquid-rust/blob/master/tests/value.rs

gnunicorn: I honestly have no idea. I don't really get the inner-workings of serde -- but it is on my bucket list :D .

I used serde_json as the base and it [defined `Null` in their value](https://docs.serde.rs/serde_json/value/enum.Value.html) and in JSON there also is no `None` but only `null`. For TOML and YAML I am not even sure a nothing-type even exists, I think it is just represented as empty strings ...

src/value.rs
gnunicorn: I am not sure, what we use the `ToString` for within templates. From the Shopify docs it is rather vague how this would come to be used. In their [cheat sheet they only explain `Nil` to be `falsy`](https://www.shopify.com/partners/shopify-cheat-sheet) and all other of their filters and loops seem to make a check for truthy-ness before executing, which probably stems from that being the implied case in ruby. And is already a behavior we don't have that way (which we discussed the other day).

I am also happy to make it `""`, which is probably closer to the interpretation of their underlying ruby implementation...

epage: You're link answered the question of what `ToString` should do

> Tags or outputs that return nil will not print anything to the page.

https://help.shopify.com/themes/liquid/basics/types#nil

btw if Shopify calls it `Nil` then we probably should too (`Value::Nil`).

epage:
Thanks for making the changes!

gnunicorn:
@epage don't be too happy to soon. I pushed before fixing the commit logs because the tests is going mad on me and wanted your input on it:

```
running 2 tests
test serialize_nil ... ok
test deserialize_nil ... FAILED

failures:

---- deserialize_nil stdout ----
	thread 'deserialize_nil' panicked at 'assertion failed: `(left == right)`
  left: `Nil`,
 right: `Nil`', tests/value.rs:79:4


failures:
    deserialize_nil

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 10 filtered out

```

Any ideas?

epage:
I wonder if `actual` is a `liquid::Str`.

epage:
I assume you'd like me to publish liquid and cobalt with this as soon as possible?

Issue #95, opened 2017-05-09
RFC: API Refactor


epage:
# Goals

Goals
- Layered so people pay for what they get
- Better separation of internals vs API
- Do not get in the way of using liquid in parallel applications
- Avoid copies with the context
  - cobalt has to copy in the vars initially
  - cobalt has reused vars that would be nice to have overlayed each other rather than copied.
- Make it possible to support expressions for include, see #142

Secondary goals
- Caching
  - Cache can be reused for cobalt's layouts/templates
    - Should this be left to clients?
  - Caching of snippets to avoid reading/parsing them
    - Is caching of top-level items good enough?

# Proposal

- [x] ParserBuilder -> Parser (#149)
  - Takes in includes, tags, blocks, and filters
- [x] Parser -> Template (#149)
  - Takes in str
- [x] Template -> String (#149)
  - Takes in globals
- [x] Separate concerns (#149)
  - Top-level Parser, Template
  - Value
  - Interpreter
  - Syntax / Compile to interpreted state
- [x] Update interpreter / syntax to take in state by reference where applicable
- [x] Update interpreter / syntax to take in state by traits
  - Including allowing custom backing stores for `globals`
  - How well can `Context` be a trait when different tags (cycle, include, break, etc) need special support in it?
- [ ] Update `trait FilterValue` to advertise the required/optional params
  - Must still support custom handling
  - Allow centralizing of error processing
  - Future: allow named parameters
- [x] Value API
  - `Value` should be an enum of `Object`, `Array`, `Scalar`, and `Nil`
  - `Scalar` is a struct wrapping an enum to hide implementation details
  - `Scalar` can coerce between different data types
- [ ] Crate structure and APIs to connect them
  - [x] liquid-value
  - [x] liquid-interpreter
  - [x] liquid-compiler
  - [x] liquid
  - [ ] tags, blocks, and filters
    - Do these sit above liquid with some API (extension traits?) to make it easy to add them to a parser?
    - Or do these sit below liquid with it having built-in support for adding these?
    - Are they added by "liquid" or "not" (like they are now) or by category with features to enable extra featurs
    - Keep in mind I want to split out from cobalt liquid-codeblock and [liquid-markdown](https://github.com/cobalt-org/cobalt.rs/issues/342)
- [ ] Should default implementations of Traits be moved out to leaf layers?
[...]

PR #141, opened 2017-10-27
Add indexed access to context values

### LINKED ISSUE #127: Bug: Array Indexes
liquid-rust version: 0.10.0
rust version: rustc 1.18.0 (03fc9d622 2017-06-06)
OS: macOS Sierra / 10.12.4

```liquid
{% assign tags = "foo,bar" | split: "," %}
tags[0]: {{ tags[0] }}
```
running the code results in

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Parser("Expected |, found [")', sr
c/libcore/result.rs:859
```

```liquid
{% for tag in tags %}
   {{ tag }}
{% endfor %}
```

Will build and display each tag. 

I've tested using another liquid [parser](http://harttle.com/shopify-liquid/) and this definitely a problem with liquid-rust.

Albibek:
This fixes #127 adding feature to access identifiers via index
- array indexes `arr[0]` are supported
- object indexes `obj["child"]` are supported and interchangeable with `obj.child`
- index expressions are not yet supported.

Checklist
- [x] `cargo test` succeeds
- [x] `rustup run nightly cargo clippy` succeeds
- [x] `cargo fmt -- --write-mode=diff` succeeds

### DIFF
--- a/src/path.rs
+++ b/src/path.rs
@@ -5,7 +5,7 @@ use token::Token;
 use token::Token::*;
 use error::{Error, Result};
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub struct IdentifierPath {
     value: String,
     indexes: Vec<Value>,
@@ -72,7 +72,7 @@ impl Renderable for IdentifierPath {
                 (value, _) if counter == 0 => Ok(value),
                 (value, _) => {
                     Error::renderer(
-                        &format!("expected indexable element, but founr '{:?}'", value)
+                        &format!("expected indexable element, but found '{:?}'", value)
                     )
                 }
             }
@@ -185,6 +185,25 @@ mod test {
     fn identifier_path_object_string() {
         let options = LiquidOptions::with_known_blocks();
         let template = "object_string: {{ test_a[0][\"test_h\"] }}\n";
+
+        let mut context = Context::new();
+        let mut internal = HashMap::new();
+        internal.insert("test_h".to_string(), Value::Num(5f32));
+
+        let test = Value::Array(vec![Value::Object(internal)]);
+        context.set_val("test_a", test);
+
+        let template = parse(template, options).unwrap();
+        assert_eq!(template.render(&mut context).unwrap(),
+                   Some("object_string: 5\n".to_owned()));
+    }
+
+    #[test]
+    #[should_panic]
+    fn identifier_path_subexpression() {
+        let options = LiquidOptions::with_known_blocks();
+        let template = concat!("{% assign somevar=\"test_h\" %}",
+                               "result_string: {{ test_a[0][somevar] }}\n");
 
         let mut context = Context::new();
         let mut internal = HashMap::new();
--- a/src/path.rs
+++ b/src/path.rs
@@ -16,47 +16,45 @@ impl Renderable for IdentifierPath {
         println!("{:?}", self);
         let value = context
             .get_val(&self.value)
-            .ok_or(Error::Render(
-                format!("{} not found in context", self.value),
-            ))?
+            .ok_or_else(|| Error::Render(format!("{} not found in context", self.value)))?
             .clone();
 
         let mut counter = self.indexes.len();
-        let result = self.indexes
-            .iter()
-            .fold(Ok(&value), |value, index| {
-                // go through error
-                let value = if let Ok(ref value) = value {
-                    value
-                } else {
-                    return value
-                };
-                counter-=1;
-       match (value, index) {
-            (&&Value::Array(ref value), &Value::Num(ref x)) => {
-                if (*x != 0f32 && !x.is_normal()) // only is_normal is not enough, because zero is not counted normal
-                    || *x < 0f32 ||
-                    x.round() > (::std::usize::MAX as f32)
-                {
-                    return Error::renderer("bad array index");
+        let result = self.indexes.iter().fold(Ok(&value), |value, index| {
+            // go through error
+            let value = if let Ok(value) = value {
+                value
+            } else {
+                return value;
+            };
+            counter -= 1;
+            match (value, index) {
+                (&Value::Array(ref value), &Value::Num(ref x)) => {
+                    // at the first condition only is_normal is not enough
+                    // because zero is not counted normal
+                    if (*x != 0f32 && !x.is_normal()) || *x < 0f32 ||
+                       x.round() > (::std::usize::MAX as f32) {
+                        return Error::renderer("bad array index");
+                    }
+                    let idx = x.round() as usize;
+                    let value =
+                        value
+                            .get(idx)
+                            .ok_or_else(|| Error::Render("index out of range".to_string()))?;
+                    Ok(value)
                 }
-                let idx = x.round() as usize;
-                let value = value.get(idx).ok_or(Error::Render(
-                    "index out of range".to_string(),
-                ))?;
-                Ok(&value)
+                (&Value::Array(_), _) => Error::renderer("bad array index type"),
+                (&Value::Object(ref value), &Value::Str(ref x)) => {
+                    let value =
+                        value
+                            .get(x)
+                            .ok_or_else(|| Error::Render("object element not found".to_string()))?;
+                    Ok(value)
+                }
+                (&Value::Object(_), _) => Error::renderer("bad object index"),
+                (value, _) if counter == 0 => Ok(value),
+                _ => Error::renderer("non-indexable element found while indexable expected"),
             }
-            (&&Value::Array(_), _) => Error::renderer("bad array index type"),
-            (&&Value::Object(ref value), &Value::Str(ref x)) => {
-                let value = value.get(x).ok_or(Error::Render(
-                    "object element not found".to_string(),
-                ))?;
-                Ok(&value)
-            }
-            (&&Value::Object(_), _) => Error::renderer("bad object index"),
-            (&value, _) if counter == 0  => Ok(value),
-            _ => Error::renderer("non-indexable element found while indexable expected")
-        }
         });
 
         match result {
@@ -121,11 +119,9 @@ mod test {
     #[test]
     fn identifier_path() {
         let options = LiquidOptions::with_known_blocks();
-        let template = concat!(
-            "array: {{ test_a[0] }}\n",
-            "complex_dot: {{ test_a[0].test_h }}\n",
-            "complex_string: {{ test_a[0][\"test_h\"] }}\n"
-        );
+        let template = concat!("array: {{ test_a[0] }}\n",
+                               "complex_dot: {{ test_a[0].test_h }}\n",
+                               "complex_string: {{ test_a[0][\"test_h\"] }}\n");
 
         let mut context = Context::new();
         let mut internal = HashMap::new();
@@ -135,15 +131,10 @@ mod test {
         context.set_val("test_a", test);
 
         let template = parse(template, options).unwrap();
-        assert_eq!(
-            template.render(&mut context).unwrap(),
-            Some(
-                concat!(
-                    "array: test_h: 5\n",
-                    "complex_dot: 5\n",
-                    "complex_string: 5\n"
-                ).to_owned(),
-            )
-        );
+        assert_eq!(template.render(&mut context).unwrap(),
+                   Some(concat!("array: test_h: 5\n",
+                                "complex_dot: 5\n",
+                                "complex_string: 5\n")
+                            .to_owned()));
     }
 }
--- a/src/path.rs
+++ b/src/path.rs
@@ -21,38 +21,60 @@ impl Renderable for IdentifierPath {
         let mut counter = self.indexes.len();
         let result = self.indexes.iter().fold(Ok(&value), |value, index| {
             // go through error
-            let value = if let Ok(value) = value {
-                value
-            } else {
-                return value;
-            };
+            let value = value?;
             counter -= 1;
             match (value, index) {
                 (&Value::Array(ref value), &Value::Num(ref x)) => {
                     // at the first condition only is_normal is not enough
                     // because zero is not counted normal
-                    if (*x != 0f32 && !x.is_normal()) || *x < 0f32 ||
-                        x.round() > (::std::usize::MAX as f32) {
-                            return Error::renderer(&format!("bad array index: '{:?}'", x));
-                        }
-                    let idx = x.round() as usize;
+                    if (*x != 0f32 && !x.is_normal()) ||
+                        x.round() > (::std::isize::MAX as f32) ||
+                            x.round() < (::std::isize::MIN as f32) {
+                                return Error::renderer(&format!("bad array index: '{:?}'", x));
+                            }
+
+                    let idx = if *x >= 0f32 {
+                        x.round() as usize
+                    } else {
+                        value.len() - (-x.round() as usize)
+                    };
+                    let err = ||
+                        Error::Render(
+                            format!("index out of range: got '{:?}' while array len is '{:?}'",
+                                    idx,
+                                    value.len()
+                                   )
+                            );
                     let value =
                         value
                         .get(idx)
-                        .ok_or_else(|| Error::Render(format!("index out of range: got '{:?}' while array len is '{:?}'", idx, value.len())))?;
+                        .ok_or_else(err)?;
                     Ok(value)
                 }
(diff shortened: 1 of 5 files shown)


src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+                return value;
+            };
+            counter -= 1;
+            match (value, index) {
+                (&Value::Array(ref value), &Value::Num(ref x)) => {
+                    // at the first condition only is_normal is not enough
+                    // because zero is not counted normal
+                    if (*x != 0f32 && !x.is_normal()) || *x < 0f32 ||
+                       x.round() > (::std::usize::MAX as f32) {
+                        return Error::renderer("bad array index");
```
epage: Could you make the error include the index?

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+            } else {
+                return value;
+            };
+            counter -= 1;
+            match (value, index) {
+                (&Value::Array(ref value), &Value::Num(ref x)) => {
+                    // at the first condition only is_normal is not enough
+                    // because zero is not counted normal
+                    if (*x != 0f32 && !x.is_normal()) || *x < 0f32 ||
+                       x.round() > (::std::usize::MAX as f32) {
```
epage: Does liquid support negative indexing like Python?  Does it support slicing?

(and this code is one of the reasons why I'm considering native integer support)

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+                        return Error::renderer("bad array index");
+                    }
+                    let idx = x.round() as usize;
+                    let value =
+                        value
+                            .get(idx)
+                            .ok_or_else(|| Error::Render("index out of range".to_string()))?;
```
epage: Can you include the index and value's length?

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+                    Ok(value)
+                }
+                (&Value::Array(_), _) => Error::renderer("bad array index type"),
```
epage: Can the error include a `{:?}` of value?

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+                (&Value::Object(ref value), &Value::Str(ref x)) => {
+                    let value =
+                        value
+                            .get(x)
+                            .ok_or_else(|| Error::Render("object element not found".to_string()))?;
+                    Ok(value)
+                }
+                (&Value::Object(_), _) => Error::renderer("bad object index"),
+                (value, _) if counter == 0 => Ok(value),
+                _ => Error::renderer("non-indexable element found while indexable expected"),
```
epage: Can you `{:?}` the element?

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+                }
+                (&Value::Array(_), _) => Error::renderer("bad array index type"),
+                (&Value::Object(ref value), &Value::Str(ref x)) => {
+                    let value =
+                        value
+                            .get(x)
+                            .ok_or_else(|| Error::Render("object element not found".to_string()))?;
+                    Ok(value)
+                }
+                (&Value::Object(_), _) => Error::renderer("bad object index"),
```
epage: Can you `{:?}` the index?

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+                            .get(idx)
+                            .ok_or_else(|| Error::Render("index out of range".to_string()))?;
+                    Ok(value)
+                }
+                (&Value::Array(_), _) => Error::renderer("bad array index type"),
+                (&Value::Object(ref value), &Value::Str(ref x)) => {
+                    let value =
+                        value
+                            .get(x)
+                            .ok_or_else(|| Error::Render("object element not found".to_string()))?;
```
epage: Can you include `x`?

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+                    Ok(value)
+                }
+                (&Value::Object(_), _) => Error::renderer("bad object index"),
+                (value, _) if counter == 0 => Ok(value),
```
epage: What is this case handling?

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+                _ => Error::renderer("non-indexable element found while indexable expected"),
+            }
+        });
+
+        match result {
+            Ok(result) => result.render(context),
+            Err(e) => Error::renderer(&format!("rendering error: {}", e)),
```
epage: You can use `chain_err` for this

```rust
result.chain_err(|| "Failed to render expression")
```

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+impl Renderable for IdentifierPath {
+    fn render(&self, context: &mut Context) -> Result<Option<String>> {
+        println!("{:?}", self);
+        let value = context
+            .get_val(&self.value)
+            .ok_or_else(|| Error::Render(format!("{} not found in context", self.value)))?
+            .clone();
+
+        let mut counter = self.indexes.len();
+        let result = self.indexes.iter().fold(Ok(&value), |value, index| {
```
epage: 1. There a reason you are using `counter` rather than `self.indexes.iter().enumerate()`?
2. [Long term](https://github.com/cobalt-org/liquid-rust/issues/95) I'm wanting to switch it so `Context` to have a trait for the user-provided variables.  Ideally, this would allow lazy lookups.  To get maximize this with indexing, the trait would have some kind of lookup object (slice of Enum(string, index).

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+    use Context;
+    use LiquidOptions;
+    use std::collections::HashMap;
+
+    #[test]
+    fn identifier_path() {
+        let options = LiquidOptions::with_known_blocks();
+        let template = concat!("array: {{ test_a[0] }}\n",
+                               "complex_dot: {{ test_a[0].test_h }}\n",
+                               "complex_string: {{ test_a[0][\"test_h\"] }}\n");
```
epage: Could you make these separate tests?

Am I reading this right that the cases can be summarized as:
- identifier_path_array_index
- identifier_path_object_dot
- identifier_path_object_index

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+        internal.insert("test_h".to_string(), Value::Num(5f32));
+
+        let test = Value::Array(vec![Value::Object(internal)]);
+        context.set_val("test_a", test);
+
+        let template = parse(template, options).unwrap();
+        assert_eq!(template.render(&mut context).unwrap(),
+                   Some(concat!("array: test_h: 5\n",
+                                "complex_dot: 5\n",
+                                "complex_string: 5\n")
```
epage: (random place)

Should we make it so all `context.get`s are IdentifierPath's?

The reason I'm considering this is we then have a central place of handling identifiers to ensure we are always consistent about how they are handled.

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+    use Renderable;
+    use Context;
+    use LiquidOptions;
+    use std::collections::HashMap;
+
+    #[test]
+    fn identifier_path() {
+        let options = LiquidOptions::with_known_blocks();
+        let template = concat!("array: {{ test_a[0] }}\n",
+                               "complex_dot: {{ test_a[0].test_h }}\n",
```
epage: is `test_a[0][var]` a failure?

src/path.rs
Albibek: Mostly other - leaf variants of Value I suppose.

src/path.rs
```diff
@@ -0,0 +1,140 @@
...
+            .clone();
+
+        let mut counter = self.indexes.len();
+        let result = self.indexes.iter().fold(Ok(&value), |value, index| {
+            // go through error
+            let value = if let Ok(value) = value {
+                value
+            } else {
+                return value;
+            };
```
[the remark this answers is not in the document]
Albibek: Nope, it's an early return from fold when value is error.
https://play.rust-lang.org/?gist=46038d66fc075362393e0eb30d9748dc&version=stable

src/path.rs
Albibek: Neither of docs I see say this explicitly, but a simple test in jekyll shows that negative indesinx is allowed, ranges(I've tried `[0..1]` `[0:1]` `[0-1]`) are not

src/path.rs
Albibek: 1. I was affraid to make iterator less readable because of more nesting.
2. `counter` is an internal variable mae to check the depth correctness. It's not used in template and is local to current function. I see no point to place it inside context.

epage: 2.  Sorry I wasn't clear.  I was referring to the whole handling of value lookups and not counter.  Both comments applied to the exact same line of code, making it harder.

Albibek: I'm sorry, but I can't understand what you mean here. Could you explain a bit more please?

epage: `Context.get_val` implements some of the parsing algorithm for looking up a value

1. Ideally we only implement the parsing algorithm in one place
2. I want to copy the minimal part of a value as possible.  Other clients wants to load data from a database (we'd change Context to have a trait to make this possible) and want to load the minimal amount from the database.

So for
```
{{ posts[5].custom.obj.arr[5] }}
```
The current implementation requires copying all of `posts` before indexing into it.

What'd be amazing is if we could do this
```rust
enum PathPart {
    ObjectIndex(&str),
    ArrayIndex(usize),
}

impl Context {
   fn get_val(&self, path: &[PathPart]) -> Option<Value> {
        ...
    }
}
```
This would allow only the leaf to be copied.

epage: Another use case for when copying will involve a lot of data: cobalt's `site.data`

Albibek: Oh, I got it. This is a complicated thing actualy. I think we could implement this in current model - where we are not using subexpressions. 
But imagine some perfect case, when we could use subexpressions, like posts[ content[2*3] ]. In this case the subexpression should be rendered with the context as a parameter first. This is a main problem Rust is protecting us from: changing the context using the value from inside the context itself. I'm not sure if introducing a trait could solve it.
I see another kind of a solution here: read-only rendering, i.e. when Context doesn't change. Probably even a separate `RenderMut` should be introduced maybe to do non-readonly rendering where it's required explicitly and which I beleive is a more rare case than immutable one.

src/path.rs
epage: Fixed

src/path.rs
epage: Whats the status of this?

src/path.rs
```diff
@@ -0,0 +1,200 @@
+use Renderable;
+use value::Value;
+use context::Context;
+use token::Token;
+use token::Token::*;
+use error::{Error, Result};
+
+#[derive(Debug)]
```
epage: Clone, Eq, etc?

src/path.rs
epage: Dropped, `?` is much better choice, the chained error wouldn't have added much value.

src/path.rs
epage: Fixed, thanks!  That'll make life much easier for people (lack of good error reporting is something we're slowly fixing).

src/path.rs
```diff
@@ -0,0 +1,200 @@
...
+                }
+                (&Value::Object(_), x) => {
+                    Error::renderer(
+                        &format!("bad object index type: expected string, but got '{:?}'", x)
+                    )
+                }
+                (value, _) if counter == 0 => Ok(value),
+                (value, _) => {
+                    Error::renderer(
+                        &format!("expected indexable element, but founr '{:?}'", value)
```
epage: founr -> found :)

src/path.rs
epage: Thanks for explaining.

I think I missed that this counts down.  So you are doing special processing on the last element.

src/path.rs
epage: Thanks for fixing this!

src/path.rs
epage: Thanks for fixing this!

src/path.rs
epage: Again, thanks!

src/path.rs
epage: Yet again, thanks!

src/path.rs
epage: More thanks

src/path.rs
epage: Thanks for looking into this!

src/path.rs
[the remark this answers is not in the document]
epage: Dropped, sorry I missed that.

src/path.rs
epage: I feel I'm missing something.  I'm not seeing why `posts[ content[2*3] ]` is a problem, particularly that Rust is protecting us from and what a read-only Context gives us.

My rough run through of how this would work:

For now, this just involves turning the `Identifier` path into `&[PathPart]` and passing that to `Context`.

Long term, when/if we add more complex lookups, we should parse the inner expression first, render it, and make a `PathPart` out of it.

This effectively transforms
```
posts[ content[2*3] ]
```
into
```
{% assign _x = 2*3 %}
{% assign _y = content[_x] %}
posts[_y]
```
(forgive my rough approximation of liquid syntax).

Note: these variables shouldn't actually need to be created in the Context, they should be able to live only in Rust.

and I don't see any problems with doing variable based lookups.  When constructing a `PathPart`, `IdentifierPath` would just do a lookup on the current value.

src/path.rs
epage: Dup of the first

epage:
So looking back, it sounds like the outstanding items are:
- Fix travis
- Fix typo while at it :)
- Respond to `is test_a[0][var] a failure?`
- (not blocking submission) Consolidate regular path parsing and indexed path parsing
- (not blocking submission) Switch to constructing a `&[PathPart]` and passing that to `Context`.

Albibek:
My current fresh rustfmt (0.9.0) refuses to format the lexer.rs as travis requires. The warning will probably will go away as Tavis updates rustfmt. I can fix it to a custom formatting if you want, but it'll be a real pain to maintain this file with newer rustfmt installed.
 
I've also added the `should_panic` test case to use when subexpressions arrive.

I'm sorry, but I currently don't have too much time to make subexpression support and I really think it should be done after grammar-based parser is implemented.

My main arguments for deferring it are (I may be wrong here, feel free to discuss it):
* I'd like to avoid writing code doing same things twice since expression based parser would make lots of code to be rewritten
* I'd like to avoid increasing amount of custom token parsing code to force ourselves for parser to happen earlier

epage:
> My current fresh rustfmt (0.9.0) refuses to format the lexer.rs as travis requires. The warning will probably will go away as Tavis updates rustfmt. I can fix it to a custom formatting if you want, but it'll be a real pain to maintain this file with newer rustfmt installed.

Huh, for some reason I( thought 0.9.0 used nightly before they decided to switch it over to rustfmt-nightly.

Would you prefer to comply with 0.8.6 first and I do the upgrade afterwords or for me to do the upgrade now, possibly causing some rebase pain for you?

> I'm sorry, but I currently don't have too much time to make subexpression support and I really think it should be done after grammar-based parser is implemented.

I'm sorry if I gave you the impression that you should!  I know it'd be complex with the current parser and I'm grateful for what you have done!

I don't mean questions to be a passive aggressive way of saying "do this" but as a means to make sure implicit assumptions become explicit so people know why you did what you did and what limitations might exist.

Albibek:
Preformed the formatting with 0.8.6.

And of course I didn't in any way imply that anyone is forcing me to do something and I'm totally no offence about this whole PR. The apology above is more about my personal feeling about work not done as it could be done if I had more time.

epage:
> And of course I didn't in any way imply that anyone is forcing me to do something and I'm totally no offence about this whole PR.

Glad to hear.  I just want to make sure I've been communicating clearly to avoid burning people out and driving away contributors.

epage:
Thank you for leaving each commit for me to see the differences between each round of feedback.

At this point, could you clean up the commit history and post here when done?
 (force pushes don't always cause email notifications)?

Thanks for getting this done!

Albibek:
Do you mean squashing commits?

epage:
> Do you mean squashing commits?

The exact format I leave to you; you know your changes best.

My recommendation
- Each commit should stand on its own
- Each commit should build cleanly (don't stress trying to test each one)

So for example, when a commit is made to satisfy the CI, it could probably be squashed.  I can go either way on improving error messages and negative indexing being separate commits or squashed, as long as the commit messages are descriptive enough.

Exceptions are gladly granted if trying to re-arrange commit history becomes messy due to conflicts.

epage:
Went ahead and did a squash+merge because I'm going to do a release soon.

Issue #155, opened 2017-12-29
Implement `contains` operator


epage:
The parser is implemented but [does nothing](https://github.com/cobalt-org/liquid-rust/blob/master/src/tags/if_block.rs#L40).

Contains should check substrings or array containment

See https://shopify.github.io/liquid/basics/operators/

PR #157, opened 2018-01-09
Implement `contains` operator

badboy:
This will panic if either side of the operator cannot be evaluated as a string.
It works for both variables and string literals as expected.

- [x] `cargo test` succeeds
- [ ] `rustup run nightly cargo clippy` succeeds
- [ ] `cargo fmt -- --write-mode=diff` succeeds <-- Tons of unrelated changes.

### DIFF
--- a/src/tags/if_block.rs
+++ b/src/tags/if_block.rs
@@ -1,5 +1,6 @@
 use error::{Error, Result};
 
+use value::Value;
 use interpreter::Argument;
 use interpreter::Context;
 use interpreter::Renderable;
@@ -25,6 +26,29 @@ struct Conditional {
     if_false: Option<Template>,
 }
 
+fn contains_check(a: &Value, b: &Value) -> Result<bool> {
+    let b = b.as_str()
+        .ok_or("Right-hand side of contains operator must be a string")?;
+
+    match *a {
+        Value::Str(ref val) => Ok(val.contains(b)),
+        Value::Object(ref obj) => Ok(obj.contains_key(b)),
+        Value::Array(ref arr) => {
+            for elem in arr {
+                let elem = elem.as_str()
+                    .ok_or("The contains operator can only check for strings")?;
+                if elem == b {
+                    return Ok(true);
+                }
+            }
+            Ok(false)
+        }
+        _ => {
+            Error::renderer("Left-hand side of contains operator must be a string, array or object")
+        }
+    }
+}
+
 impl Conditional {
     fn compare(&self, context: &Context) -> Result<bool> {
         let a = self.condition.lh.evaluate(context)?;
@@ -37,7 +61,7 @@ impl Conditional {
             ComparisonOperator::GreaterThan => a > b,
             ComparisonOperator::LessThanEquals => a <= b,
             ComparisonOperator::GreaterThanEquals => a >= b,
-            ComparisonOperator::Contains => false, // TODO!!!
+            ComparisonOperator::Contains => contains_check(&a, &b)?,
         };
 
         Ok(result == self.mode)
@@ -133,6 +157,7 @@ pub fn if_block(_tag_name: &str,
 
 #[cfg(test)]
 mod test {
+    use std::collections::HashMap;
     use super::*;
     use value::Value;
     use compiler;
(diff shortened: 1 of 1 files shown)


epage:
For reference, #155

src/tags/if_block.rs
```diff
@@ -25,6 +26,19 @@ struct Conditional {
...
+    let a = match a.as_str() {
+        Some(s) => s,
+        None => return Error::renderer("Left-hand side of contains operator must be a string"),
+    };
+    let b = match b.as_str() {
+        Some(s) => s,
+        None => return Error::renderer("Right-hand side of contains operator must be a string"),
+    };
+
+    Ok(a.contains(b))
```
epage: `contains` supports two use cases
- substring matches (`page.title contains "Hello"`)
- checking a hashmap if it contains a key (`page contains "title"`)

This implements the first but for https://github.com/cobalt-org/cobalt.rs/issues/363, it sounds like you need the second.

If you want, I can merge this now.  Hopefully soon I can add support for checking for the object containing a key.  Otherwise, feel free to add support for that in this PR or another one.  Just let me know what you want.

src/tags/if_block.rs
```diff
@@ -25,6 +26,19 @@ struct Conditional {
     if_false: Option<Template>,
 }
 
+fn contains_check(a: Value, b: Value) -> Result<bool> {
+    let a = match a.as_str() {
+        Some(s) => s,
+        None => return Error::renderer("Left-hand side of contains operator must be a string"),
```
epage: (this is non-essential feedback, feel free to ignore)

Personally, I prefer
```rust
let a = a.as_str().ok_or("Left-hand side of contains operator must be a string")?;
```
or more like yours
```rust
let a = a.as_str().ok_or_else(|| Error::renderer("Left-hand side of contains operator must be a string"))?;
```

src/tags/if_block.rs
```diff
@@ -336,4 +350,75 @@ mod test {
...
+    #[should_panic(expected = "Left-hand side")]
+    fn contains_comparison_non_string_lhs() {
+        let text = "{% if 7 contains \"Star\" %}if true{% endif %}";
+        let tokens = compiler::tokenize(&text).unwrap();
+        let template = compiler::parse(&tokens, &options())
+            .map(interpreter::Template::new)
+            .unwrap();
+
+        let mut context = Context::new();
+        let _output = template.render(&mut context).unwrap();
```
epage: Instead of checking for panic, you could

```rust
assert!(output.is_err());
```
This makes maintenance a little easier because you don't have to worry about string matching the panic message (the error message can change without breaking the tests).

(also applies to the next test in case you decide to implement it)

src/tags/if_block.rs
```diff
@@ -336,4 +350,75 @@ mod test {
         let output = template.render(&mut context).unwrap();
         assert_eq!(output, Some("fourth".to_owned()));
     }
+
+    #[test]
+    fn contains_comparison() {
```
epage: Personally, I prefer the tests to be broken up by their purpose.

This is helpful for the person modifying tests and the person analyzing a test failure.

So for example,
```rust
#[test]
fn contains_string_literals() {
   ...
}

#[test]
fn contains_else() {
   ...
}

#[test]
fn test contains_variable() {
   ...
}
```

(I don't recommend literally copying my function names above when splitting this function because I'm making guesses as to your intent.  You know what you intended with each test and this serves as a way for you to document it)

badboy:
Thanks for quickly reviewing this. This was just a first shot at the minimal implementation.
Will follow your advice and split up tests and also implement the HashMap `contains`

badboy:
1. Split up tests
2. Implemented for objects and arrays
3. Cleaned up extensive `match`es

src/tags/if_block.rs
```diff
@@ -25,6 +26,25 @@ struct Conditional {
     if_false: Option<Template>,
 }
 
+fn contains_check(a: &Value, b: &Value) -> Result<bool> {
+    let b = b.as_str().ok_or("Right-hand side of contains operator must be a string")?;
+
+    match *a {
+        Value::Str(ref val) => Ok(val.contains(b)),
+        Value::Object(ref obj) => Ok(obj.contains_key(b)),
+        Value::Array(ref arr) => {
```
epage: I looked up `contains` but I thought it said it works for substrings and key checks.

Looking it up, instead its documented to work for substrings and array membership.
https://shopify.github.io/liquid/basics/operators/

But the implementation does work for objects's
- https://github.com/Shopify/liquid/blob/master/lib/liquid/condition.rb#L18
- https://ruby-doc.org/core-2.4.2/Hash.html#method-i-include-3F

So the only thing we're missing, compared to official liquid, is converting the right side to a string if the left side is a string.

That I'm perfectly happy postponing.  I've started a major revamp for `Value` that will make it easier to implement all of these random liquid conversions / coercions that take place.

epage:
This looks great!  The work I see remaining to do
- Resolve any CI failures
- Squash commits
- Please title the commit something like `feat(if): Support contains operator`
- Please reference that this `Fixes #155`

Again, if there is any of that you aren't comfortable with or want to hand off, feel free to do so.

(Sorry we're using an odd version of rustfmt.  I tried to mitigate that with writing out its config)

badboy:
Ah no problem. With these detailed instructions it will be easy to do.

epage:
Huh, I should actually give the "why" to things.

I use clog-cli to create changelists.  By having `feat: stuf` in the title of the commit, I won't miss announcing your new feature in the changelog.

`Fixes #155` will auto-close the issue and link it to the commit.

Squashing  is to make the history cleaner which makes it easier to browse or `git bisect`.

badboy:
> I use clog-cli to create changelists. By having feat: stuf in the title of the commit, I won't miss announcing your new feature in the changelog.

Yup, I do something similar in other projects.
The rest is known as well. Thanks!

badboy:
So I`m a bit at a loss here
rustfmt currently trips on this:
```patch
-        let text =
-            "{% if \"Star Wars\" contains \"Star Trek\"  %}if true{% else %}if false{% endif %}";
+        let text = "{% if \"Star Wars\" contains \"Star Trek\"  %}if true{% else %}if false{% endif %}";
```

If I used that

badboy:
Oops, managed to hit the wrong button.

badboy:
The only remaining failure looks like a network error on traivs, restarting that test case might resolve it.

epage:
Awesome! I kicked off the jenkins job.

I feel this will justify a liquid and cobalt release to get this into yours (and others) hands.

epage:
btw the crate ecosystem seems to be trying to standardize the license.  Please look at https://github.com/cobalt-org/liquid-rust/issues/7 and leave a comment if you accept the change.

epage:
`cargo publish` on cobalt v0.11.1 is happening right now

Issue #88, opened 2017-04-29
`divided_by` filter incorrectly rounds down when divisor is float


kennytm:
[According to documentation](https://shopify.github.io/liquid/filters/divided_by/), `divided_by` floors only when the divisor is an integer. Otherwise it should not round.

```
{{ 20 | divided_by: 7 }}
⇒ 2

{{ 20 | divided_by: 7.0 }}
⇒ 2.857142857142857

{% assign my_integer = 7 %}
{{ 20 | divided_by: my_integer }}
⇒ 2

{% assign my_integer = 7 %}
{% assign my_float = my_integer | times: 1.0 %}
{{ 20 | divided_by: my_float }}
⇒ 2.857142857142857
```

However, as of 0.9.1, all of these produce `2` as result since [it floors unconditionally](https://github.com/cobalt-org/liquid-rust/blob/73d480075a70f40028cdb41034ae601cf8366e70/src/filters.rs#L221).

I think you may need to split `Value::Num` into two distinct variants to properly fix it.

epage:
Thanks for catching this.

Yeah, this will probably require integer support and we should probably audit the rest of the filters for when we should be returning integer or decimal values.

epage:
Fixed by #158

PR #158, opened 2018-01-14
Re-work `Value` to "final" API, cleaning up problems along the way

epage:
This ended up causing significant performance improvement.  It could have been more but for some reason the initial commit slowed down variables for some reason.


epage:
My reference to fixing #99 is supposed to be #88

Issue #138, opened 2017-10-23
RFC: Use a non-regex parser


epage:
This is a possible enhancement that could make it easier to address some of our bugs and missing features.

The downside is that shopofy/liquid uses regex iirc so it'll be more difficult to follow their grammar.  I have heard that there are some differences in the regex engines that already cause some problems.

### Goals
- Performance (faster than what we have now at least)
- Making it easier to implement features, for example
  - More complex indexing (https://github.com/cobalt-org/liquid-rust/issues/145)
  - Variable includes (https://github.com/cobalt-org/liquid-rust/issues/142)
  - Named arguments for filters (https://github.com/cobalt-org/liquid-rust/issues/92, also requires changes to filter API)
  - nil literals (https://github.com/cobalt-org/liquid-rust/issues/223)
  - blank / empty keywords (https://github.com/cobalt-org/liquid-rust/issues/222)
- Generally easier to maintain
- A concise top-level that reduces churn (all helpers can be moved into a sub-crate and break compat whenever)
- File location information in errors (formerly #232)

### Background

There are really to main issues with the current parser
- The performance of the lexer (I'm assuming thats whats so slow)
- The maintainability of the parser

The parser right now is cobbled together functions that have too much logic in them, causing others to duplicate it.  For example, the for-block has named arguments but not filters, because we don't have reusable argument parsing.  The parser is also inconsistent on how it consumes tokens.  A consistent, composible model would be a big help.

# Solutions

## Patch up the current approach.

The parser could be made more maintainable while keeping the current token stream if desired.  We could make combinators for token processing, that return where they left off parsing (maybe even on failure like nom's IResult).  We can then refactor all of our parsers into using these combinators so we can have an easier to maintain code base.

# Parser comparison

- nom
- [Combine](https://github.com/Marwes/combine)
  - [Author moved on to lalrop](https://www.reddit.com/r/rust/comments/6w6vak/writing_a_pl_in_rust_lalrpop_or_nom_or/dm5ttgh/)
- [Chomp](https://github.com/m4rw3r/chomp)
- pest
- [lalrop](https://github.com/lalrpop/lalrpop)
- [pom](https://github.com/J-F-Liu/pom)

## Benchmarks

- https://github.com/Geal/parser_benchmarks
- https://m4rw3r.github.io/parser-combinator-experiments-rust
- https://m4rw3r.github.io/parser-combinator-experiments-errors
- https://m4rw3r.github.io/parser-combinator-experiments-part-3
[...]

Issue #114, opened 2017-05-25
Stabalize the Error type


epage:
As the [guidelines](https://github.com/brson/rust-api-guidelines#c-stable) point out, error types are a way to leak out unintended dependencies in a public API, possibly breaking people as you evolve it.

Types of errors
- compiler
- interpret
- dependency
  - third party
  - compiler plugin error
  - interpreter plugin error
  - filter plugin error

Programmaticaly relevant information
- high level error type
- displayable error message
- displayable API backtrace
- displayable user backtrace

epage:
#105 might also have impact on stabilizing the error type.

epage:
Valuable things to consider
- chaining context (like location, filter, values, etc)
- plugin errors and being able to chain them in with the regular errors

epage:
Main error type
- Lexer(String)
- Parser(String)
- Render(String)
- Filter
- Other(String)
- Io

FilterError
- InvalidType(String)
- InvalidArgumentCount(String)
- InvalidArgument(position, String)

epage:
Closed by #164

Issue #105, opened 2017-05-20
Display location information for errors


epage:
> It would be much nicer if instead of [error] Liquid error: Rendering error: Tried to iterate over None, which is not supported. or something similarly unhelpful, cobalt would directly point to the problem.

This is for the liquid side of the work for cobalt-org/cobalt.rs#136

PR #164, opened 2018-01-20
Provide context to errors

epage:
This effectively addresses #105.  Anything further will be in #138.

The error API is mostly in shape now (#114).  `FilterError` still needs some work but I want to first see what help we provide people as we refactor how filters are created.


epage:
Example of a parse error:
```
Error: parse error: Snippet does not exist
cause: No such file or directory (os error 2)
  with:
    non-existent source=""
from: {% include blue %}
from: {% for i in (1..max) reversed %}
from: {% if 5 == var %}
```

Example of a render error:
```
Error: liquid: Expected whole number, found `fractional number`
  with:
    end=10.5
from: {% for i in (1..max) reversed %}
from: {% if blue skies == var %}
```

Geobert:
Do we have the file and line number as well?

epage:
> Do we have the file and line number as well?

Short answer, no.  This is at least providing some context which is why I'm closing out #105 with this.  We might as well get line numbers when we do #138.

PR #180, opened 2018-07-20
Upgrade to f64 floats

jjl:
BREAKING CHANGE: Liquid's "Value" API now works with f64's instead of f32's
- [ ] `cargo test` succeeds
- [ ] `rustup run nightly cargo clippy` succeeds
- [ ] `cargo fmt -- --write-mode=diff` succeeds

### DIFF
--- a/src/filters/math.rs
+++ b/src/filters/math.rs
@@ -167,9 +167,9 @@ mod tests {
 
     #[test]
     fn unit_abs() {
-        let input = Value::scalar(-1f32);
+        let input = Value::scalar(-1f64);
         let args = &[];
-        let desired_result = Value::scalar(1f32);
+        let desired_result = Value::scalar(1f64);
         assert_eq!(unit!(abs, input, args), desired_result);
     }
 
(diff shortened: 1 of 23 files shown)


src/filters/math.rs
```diff
@@ -247,32 +247,32 @@ mod tests {
...
         assert_eq!(
-            unit!(modulo, Value::scalar(24_f32), &[Value::scalar(7_f32)]),
-            Value::scalar(3_f32)
+            unit!(modulo, Value::scalar(24_f64), &[Value::scalar(7_f64)]),
+            Value::scalar(3_f64)
         );
         assert_eq!(
-            unit!(modulo, Value::scalar(183.357), &[Value::scalar(12_f32)]),
+            unit!(modulo, Value::scalar(183.357), &[Value::scalar(12_f64)]),
             Value::scalar(3.3569946)
```
epage: > Not sure what's causing the failing float test - any ideas?

I'm assuming we were rounding due to f32 and now with f64, we have more precision.  IUn theory, we just change the result we are comparing aginst.

epage:
1. Could you change the title of your commit to "feat: Upgrade from f32 to f64"
2. Could you add the following section

> BREAKING CHANGE: Liquid's "Value" API now works with f64's instead of f32's

This will help me remember to bump semver

jjl:
I've changed the commit message, not sure where you want this section adding - i've added it to the PR message

epage:
> I've changed the commit message, not sure where you want this section adding - i've added it to the PR message

At the bottom of your commit message please.

For more info, see https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular

PR #192, opened 2018-09-05
perf(render): Use a Write

### LINKED ISSUE #187: Render to `Writer`
From @emoon

Also the ability to to do `Template::render_to<W: Write>(&mut dest: W, ...)` would be useful as well as now I always have to copy the resulting data to my destination (while Template would have been able to write to it directly, removing the copy and memory allocation)

epage:
`render_to` uses an `io::Write`.  This is hopefully better than string
concatenation, especially if you can write directly to file. `render` is
still available for convinience (though it no longer returns an
`Option`).

Fixes #187

BREAKING CHANGES: Changed focus to `io::Write`
- `Renderable`s must implement `render_to`.
- `render` no longer returns an `Option`.

### DIFF
--- a/src/compiler/parser.rs
+++ b/src/compiler/parser.rs
@@ -404,6 +404,9 @@ mod test_expect {
 
 #[cfg(test)]
 mod test_split_block {
+    use std::collections::HashMap;
+    use std::io::Write;
+
     use super::super::split_block;
     use super::super::tokenize;
     use super::super::BoxedBlockParser;
@@ -412,14 +415,13 @@ mod test_split_block {
     use interpreter;
     use interpreter::Context;
     use interpreter::Renderable;
-    use std::collections::HashMap;
 
     #[derive(Debug)]
     struct NullBlock;
 
     impl Renderable for NullBlock {
-        fn render(&self, _context: &mut Context) -> Result<Option<String>> {
-            Ok(None)
+        fn render_to(&self, _writer: &mut Write, _context: &mut Context) -> Result<()> {
+            Ok(())
         }
     }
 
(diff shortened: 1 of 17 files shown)


epage:
Before:

test bench_parse_template  ... bench:      25,826 ns/iter (+/- 1,689)
test bench_parse_text      ... bench:         634 ns/iter (+/- 52)
test bench_parse_variable  ... bench:       2,527 ns/iter (+/- 182)
test bench_render_template ... bench:       7,225 ns/iter (+/- 2,762)
test bench_render_text     ... bench:         262 ns/iter (+/- 20)
test bench_render_variable ... bench:       1,392 ns/iter (+/- 112)

epage:
After

test bench_parse_template  ... bench:      25,657 ns/iter (+/- 2,238)
test bench_parse_text      ... bench:         629 ns/iter (+/- 37)
test bench_parse_variable  ... bench:       2,562 ns/iter (+/- 203)
test bench_render_template ... bench:       7,600 ns/iter (+/- 726)
test bench_render_text     ... bench:         248 ns/iter (+/- 105)
test bench_render_variable ... bench:       1,368 ns/iter (+/- 101)

Note: no real allocation change was made to the benchmark.  It still calls `render` rather than `render_to`.  In addition, it doesn't have many allocation heavy code paths.

emoon:
Nice. Did you consider taking a generic parameter for Write? Now it's a trait object so that adds some extra overhead but perhaps it doesn't affect that much.

epage:
It wouldn't have worked with how we track Renderables.  They are boxed, so we need an object-safe function which means `dyn Write` rather than `impl Write`.

emoon:
Ah, I see

PR #211, opened 2018-10-15
Additional shopify tags

### LINKED ISSUE #163: Additional shopify tags
The official liquid ruby gem has the following additional tags
- [ ] [increment](https://shopify.github.io/liquid/tags/variable/)
- [ ] [decrement](https://shopify.github.io/liquid/tags/variable/)
- [ ] ifchanged
- [ ] [tablerow](https://shopify.github.io/liquid/tags/iteration/)

See also https://github.com/Shopify/liquid/tree/master/lib/liquid/tags

Goncalerta:
Resolves #163
Tags to implement:
- [x] increment
- [x] decrement
- [x] ifchanged
- [x] tablerow


Goncalerta:
I'm a little bit confused about the use of the `trace_with` function and about where exactly to call it on the increment/decrement  `render_to` function (or if it is even needed at all).

epage:
> I'm a little bit confused about the use of the trace_with function and about where exactly to call it on the increment/decrement  render_to function (or if it is even needed at all).

The intention is to create a liquid backtrace when errors are reported.

There are two main times we do this
- When parsing a block, add a backtrace line on any error from the block's contents (ie the stuff between `{% tag %}` and `{% endtag %}`.
- When rendering a tag/block, add a backtrace line to any error created.

epage:
Overall, this looks great!

Goncalerta:
> > I'm a little bit confused about the use of the trace_with function and about where exactly to call it on the increment/decrement  render_to function (or if it is even needed at all).
> 
> The intention is to create a liquid backtrace when errors are reported.
> 
> There are two main times we do this
> 
> * When parsing a block, add a backtrace line on any error from the block's contents (ie the stuff between `{% tag %}` and `{% endtag %}`.
> * When rendering a tag/block, add a backtrace line to any error created.

So, I suppose in this case it should be added after the `chain` method?

epage:
> So, I suppose in this case it should be added after the chain method?

Technically, yes but the likelihood of it failing is so low, [apparently I didn't bother adding traces to `render_to` when I created it](https://github.com/cobalt-org/liquid-rust/blob/master/src/tags/cycle_tag.rs#L34)

> I tested that, and I found out the method only works for liquid errors, so it can't be used there. So does that mean in this particular case there is no need for using that function?

Wait, doesn't `chain` make it into a `liquid::Error`, making it able to be used?

Goncalerta:
> Technically, yes but the likelihood of it failing is so low, [apparently I didn't bother adding traces to `render_to` when I created it](https://github.com/cobalt-org/liquid-rust/blob/master/src/tags/cycle_tag.rs#L34)

Ok, so I suppose i will just get rid of the trace functions.

> Wait, doesn't `chain` make it into a `liquid::Error`, making it able to be used?

Oh, sorry, I must have glanced at the error message too fast (or perhaps vs code took too much time to update from the last error message, it sometimes does that to me) because now that I'm trying to reproduce the error, it actually is a `liquid::Error`. The only error I get now is about not importing the trait, which could be easily fixable.

Goncalerta:
While working on the `tablerow` tag, I've noticed that the ranges in the for loop aparently work different than specified in the Liquid tutorial.

This liquid code (present in the [range section of the for loop parameters](https://shopify.github.io/liquid/tags/iteration/)):
```
{% for i in (3..5) %}
  {{ i }}
{% endfor %}

{% assign num = 4 %}
{% for i in (1..num) %}
  {{ i }}
{% endfor %}
```
supposedly should output 
```
3 4 5
1 2 3 4
```
But in this crate it actually outputs
```
 3 4
 1 2 3
```
Is there any reason for the ranges here to be exclusive? Or is it a bug?

epage:
> Is there any reason for the ranges here to be exclusive? Or is it a bug?

I'd consider that a bug.  Could you fix that in a distinct PR?

epage:
Finally got a chance to dig through tablerow.  It also is looking great!

src/tags/ifchanged_block.rs
```diff
@@ -238,6 +57,7 @@ mod test {
     use value::Object;
     use value::Value;
 
+    /*
```
epage: Don't forget to clean this up./

liquid-interpreter/src/context.rs
```diff
@@ -328,6 +348,11 @@ impl<'g> Context<'g> {
         CycleState { context: self }
     }
 
+    /// Access the block's `IfChangedState`.
+    pub fn ifchanged(&mut self) -> &mut IfChangedState {
```
epage: I need to look into a `registers` thing like ruby liquid to avoid having to add tag-specific knowledge to the `Context`.

Goncalerta:
Ok, now I think that after merging #212 and fixing the tablerow unit tests everything will be working fine regarding these four tags.

If I'm missing something please tell me.

epage:
Everything looks great.  Feel free to clean up your commit history and let me know (github doesn't notify on force-pushes) and I'll merge this.

Goncalerta:
@epage Ok, I thing everything is set now. Let's see if it passes all tests.

PR #212, opened 2018-10-16
fix(for_block): make ranges inclusive.

Goncalerta:
For loops with ranges should include the last value.


epage:
Wondering how this came about, it appears it has been this way [since the beginning](https://github.com/cobalt-org/liquid-rust/commit/8fdcac40840d30e90942382ae36cc99443e2a80b#diff-b213da8ac4f7fdd7b5ac45e00ae854f0R67).  I've not seen anything to indicate this being an intentional deviation.

epage:
btw all looks good and will merge once tests pass

PR #213, opened 2018-10-17
feat(for_block): support parameters with variables

### LINKED ISSUE #162: `{% for %}`'s `offset` and `limit` should support variables
See https://github.com/Shopify/liquid/blob/master/lib/liquid/tags/for.rb#L122

Goncalerta:
Make forloops and tablerows support variables in their parameters (`offset`, `limit`, `cols`).

Closes #162

### DIFF
--- a/src/tags/for_block.rs
+++ b/src/tags/for_block.rs
@@ -6,6 +6,7 @@ use itertools;
 use liquid_error::{Result, ResultLiquidChainExt, ResultLiquidExt};
 use liquid_value::{Object, Scalar, Value};
 
+use compiler::value_token;
 use compiler::Element;
 use compiler::LiquidOptions;
 use compiler::Token;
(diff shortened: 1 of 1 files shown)


src/tags/for_block.rs
```diff
@@ -98,14 +99,41 @@ fn iter_slice(
     slice
 }
 
+/// Extracts an integer value or an identifier from the token stream
+fn parse_attr(args: &mut Iter<Token>) -> Result<Argument> {
+    expect(args, &Token::Colon)?;
+    match args.next() {
+        t @ Some(Token::IntegerLiteral(_)) | t @ Some(Token::Identifier(_)) => {
+            value_token(t.unwrap().clone())?.to_arg()
```
epage: Added a note to https://github.com/cobalt-org/liquid-rust/issues/95 to make `value_token` work by reference to avoid all these unnecessary clones (this and elsewhere)

PR #215, opened 2018-10-19
feat(tablerow): add tablerow object for its tag

Goncalerta:
Similar to forloop object, but for tablerow.

### DIFF
--- a/src/tags/for_block.rs
+++ b/src/tags/for_block.rs
@@ -377,14 +377,6 @@ fn trace_tablerow_tag(
     )
 }
 
-fn position_in_row(index: usize, columns_per_row: Option<usize>) -> usize {
-    if let Some(columns_per_row) = columns_per_row {
-        index % columns_per_row
-    } else {
-        index
-    }
-}
-
 impl Renderable for TableRow {
     fn render_to(&self, writer: &mut Write, context: &mut Context) -> Result<()> {
         let mut range = self.range.evaluate(context).trace_with(|| self.trace())?;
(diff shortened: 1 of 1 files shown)


src/tags/for_block.rs
```diff
@@ -394,16 +386,42 @@ impl Renderable for TableRow {
...
+                let col_first = col_index == 0;
+                let col_last = cols.filter(|&cols| col_index + 1 == cols).is_some() || last;
+
+                if col_first {
+                    write!(writer, "<tr class=\"row{}\">", row_index + 1)
+                        .chain("Failed to render")?;
                 }
-                col_num += 1;
-                write!(writer, "<td class=\"col{}\">", col_num).chain("Failed to render")?;
+                write!(writer, "<td class=\"col{}\">", col_index + 1).chain("Failed to render")?;
```
epage: I think I'd prefer `helper_vars` to be initialized first.  That leaves it so we have all the writes in a row, making it easier to see whats happening.

If for-loop isn't doing this, then oops (granted, I didn't write it)

src/tags/for_block.rs
```diff
@@ -394,16 +386,42 @@ impl Renderable for TableRow {
...
+
+                helper_vars.insert("index0".into(), Value::scalar(i as i32));
+                helper_vars.insert("index".into(), Value::scalar((i + 1) as i32));
+                helper_vars.insert(
+                    "rindex0".into(),
+                    Value::scalar((range.len() - i - 1) as i32),
+                );
+                helper_vars.insert("rindex".into(), Value::scalar((range.len() - i) as i32));
```
epage: Are the variables documented anywhere?

epage:
Thanks!

PR #219, opened 2018-10-22
 feat(value): dot notation form of size for arrays and objects

### LINKED ISSUE #136: Dot notation form of size not implemented?
**liquid-rust version:** 0.10.1
**rust version:** rustc 1.21.0 (3b72af97e 2017-10-09)
**OS:** Arch Linux

I'm using Cobalt (0.7.3) and it appears that the [dot notation form of size](https://shopify.github.io/liquid/filters/size/) doesn't work.

With the following in `index.liquid` and 10 posts:

```
{{ posts | size }}
x
{{ posts.size }}
```
the output is:

```
10
x
```

What I'm trying to do is show some text if there are more than 5 posts:

```
{% if posts.size > 5 %}
  more posts
{% endif %}
```

Goncalerta:
Resolves #136

My proposal is to implement this feature in `Variable::evaluate` instead of `Value::get`. I find that this makes more sense because when we use `.size` we are actually _evaluating_ the size of a value (as in, creating a new `Value`) instead of retrieving a value from inside another value. On the other hand, this avoids changing the signature of `Value::get`, to return actual `Value`s instead of references, which would imply changing the code in many other places.

Another change that is tightly related to this feature is refactoring `Variable::evaluate` to return ` Result<Value>` instead of `Result<Path>`. This also makes more sense in my head because evaluating seems to imply calculating the value held in the variable, while resolving the path to the value seems to be just the first step to do that. On the other hand, that avoids code duplication regarding the `.size` feature.

### DIFF
--- a/liquid-interpreter/src/expression.rs
+++ b/liquid-interpreter/src/expression.rs
@@ -26,10 +26,7 @@ impl Expression {
     pub fn evaluate(&self, context: &Context) -> Result<Value> {
         let val = match *self {
             Expression::Literal(ref x) => x.clone(),
-            Expression::Variable(ref x) => {
-                let path = x.evaluate(context)?;
-                context.stack().get(&path)?.clone()
-            }
+            Expression::Variable(ref x) => x.evaluate(context)?,
         };
         Ok(val)
     }
--- a/liquid-interpreter/src/variable.rs
+++ b/liquid-interpreter/src/variable.rs
@@ -4,6 +4,7 @@ use std::io::Write;
 use error::{Error, Result, ResultLiquidChainExt};
 use value::Path;
 use value::Scalar;
+use value::Value;
 
 use super::Context;
 use super::Expression;
@@ -30,7 +31,7 @@ impl Variable {
     }
 
     /// Convert to a `Path`.
-    pub fn evaluate(&self, context: &Context) -> Result<Path> {
+    pub fn evaluate(&self, context: &Context) -> Result<Value> {
         let path: Result<Path> = self
             .path
             .iter()
@@ -43,7 +44,11 @@ impl Variable {
                     .clone();
                 Ok(s)
             }).collect();
-        path
+        let path = path?;
+
+        let value = context.stack().get(&path)?.clone();
+
+        Ok(value)
     }
 }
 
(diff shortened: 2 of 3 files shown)


epage:
> My proposal is to implement this feature in Variable::evaluate instead of Value::get. I find that this makes more sense because when we use .size we are actually evaluating the size of a value (as in, creating a new Value) instead of retrieving a value from inside another value.

I think I prefer all value-lookup logic to be centralized.  One of the reasons is that I want to expand the `Globals` API to have full path lookup.  I want to make it so Rust structs can implement the trait, allowing Liquid to get a view of Rust data with the minimal amount of data conversion.  Keeping `size` in `Value`, it allows this to also be deferred into the `Globals` trait implementation as well.

RE Mental model: If you think of it as an immutable data structure (which they are) then its less clear if a `size` is really a calculation being evaluated or not.

So yes, I've been planning on switching things to `Cow` to allow for non-reference to be returned.  I was in the middle of all of this  (#95) but took a break to get features out to a cobalt user.

> Another change that is tightly related to this feature is refactoring Variable::evaluate to return Result<Value> instead of Result<Path>. This also makes more sense in my head (I hope I'm not missing the reasoning over it) because evaluating seems to imply calculating the value held in the variable, while resolving the path to the value seems to be just the first step to do that. On the other hand, that avoids code duplication regarding the .size feature.

Yeah, its confusing with the current name.  My expectation is to rename `evaluate` to `resolve`.  If it works, we could also provide an `evaluate`.

Goncalerta:
One more question, 

> Yeah, its confusing with the current name. My expectation is to rename `evaluate` to `resolve`. If it works, we could also provide an `evaluate`.

Is there any advantage in having a separate `resolve` function to calculate the path instead of doing the full `evaluate` and returning a Value? I'm just asking out of curiosity.

epage:
> Is there any advantage in having a separate resolve function to calculate the path instead of doing the full evaluate and returning a Value? I'm just asking out of curiosity.

I'm unsure of any technical value at the moment.  Its mostly separation of concerns.  A `Variable` is an un-evaluated reference to data, a `Path` is the evaluated version.  From that, I made `Variable` responsible for only the part it needed, allowing the users of that API to do with it as they needed.

PR #220, opened 2018-10-27
feat(filters): array manipulation filters

Goncalerta:
Implemented some jekyll filters related to arrays:
- push
- pop
- unshift
- shift
- array_to_sentence_string


epage:
Related issue: https://github.com/cobalt-org/cobalt.rs/issues/350

src/filters/array.rs
```diff
@@ -0,0 +1,171 @@
+use liquid_value::Value;
+
+use super::{check_args_len, invalid_input};
+use interpreter::FilterResult;
+
+/// Receives a `Value::Array` as an input.
+/// Returns a copy of the input with the given value appended at the end.
+pub fn push(input: &Value, args: &[Value]) -> FilterResult {
```
epage: Since this from jekyll and not liquid, we should put these under `extra-filters`

Granted, we need to re-look into how to handle that at some point

epage:
Sorry, didn't get to finish.  I'll try to get back to this later toda

src/filters/array.rs
```diff
@@ -0,0 +1,171 @@
...
+        let desired_result = Value::Array(vec![
+            Value::scalar("Seattle"),
+            Value::scalar("Tacoma"),
+            Value::scalar("Spokane"),
+        ]);
+        assert_eq!(unit!(push, input, args), desired_result);
+    }
+
+    #[test]
+    fn unit_pop() {
```
epage: We should test pop on an empty array

src/filters/array.rs
```diff
@@ -0,0 +1,171 @@
...
+        let desired_result = Value::Array(vec![
+            Value::scalar("Olympia"),
+            Value::scalar("Seattle"),
+            Value::scalar("Tacoma"),
+        ]);
+        assert_eq!(unit!(unshift, input, args), desired_result);
+    }
+
+    #[test]
+    fn unit_shift() {
```
epage: we should also test shift on an empty array

src/filters/array.rs
```diff
@@ -0,0 +1,171 @@
...
+        let input = Value::Array(vec![Value::scalar("Seattle"), Value::scalar("Tacoma")]);
+        let args = &[];
+        let desired_result = Value::Array(vec![Value::scalar("Tacoma")]);
+        assert_eq!(unit!(shift, input, args), desired_result);
+    }
+
+    #[test]
+    fn unit_array_to_sentence_string() {
```
epage: We should test 0 and 1 items

src/filters/array.rs
epage: Sorry, didn't catch that you did this in the mod.rs

src/filters/array.rs
```diff
@@ -0,0 +1,171 @@
...
+        .ok_or_else(|| invalid_input("Array expected"))?
+        .iter();
+
+    let len = array.len();
+    match len {
+        0 => Ok(Value::scalar("")),
+        1 => Ok(Value::scalar(array.next().unwrap().to_string())),
+        _ => {
+            let first = array.next().unwrap().to_string();
+            let last = array.next_back().unwrap().to_string();
```
epage: Wouldn't `to_str` be preferred since we don't an owned version

And granted, while performance for this isn't as critical, you could take it a step further and `write!` into a buffer for this entire operation.  That'll remove all of the allocations from `to_string` and `format` (O(N) allocations) and instead have the allocations of the `Vec` growing which is usually optimized pretty well.

Doing this, you *might* even be able to merge the code paths or 0, 1, and n size arrays.

Goncalerta:
I'm just unsure if unwraping  `write!` is the way to go. As far as I'm aware, it should be safe to write to a string but if I'm mistaken I can make it return an error instead.

src/filters/array.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    #[test]
+    fn unit_array_to_sentence_string() {
+        let input = Value::Array(vec![
+            Value::scalar("foo"),
+            Value::scalar("bar"),
+            Value::scalar("baz"),
+        ]);
+        let args = &[];
+        let desired_result = Value::scalar("foo, bar and baz");
```
epage: 1. Does jekyll's version leave off the oxford comma?

With oxford comma

> foo, bar, and baz

Without

> foo, bar and baz

2. Should we follow them?

src/filters/array.rs
```diff
@@ -0,0 +1,217 @@
...
+    Ok(Value::array(array))
+}
+
+/// Convert an array into a sentence.
+pub fn array_to_sentence_string(input: &Value, args: &[Value]) -> FilterResult {
+    check_args_len(args, 0, 1)?;
+
+    let connector = args
+        .get(0)
+        .and_then(|arg| Some(arg.to_string()))
```
epage: Sorry I missed this earlier.  Does this need to be `to_string` or can it be `to_str`?

For example:
```rust
let possible_connector = args
    .get(0)
    .and_then(|arg| Some(arg.to_str()));
let connector = possible_connector
    .map(|s| s.as_ref)
    .unwrap_or_else("and");
```
- `possible_connector` would be a `Option<Cow<'static, str>>`
- `connector` would be a `&str`

src/filters/array.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    let mut sentence = array
+        .next()
+        .and_then(|v| Some(v.to_string()))
+        .unwrap_or_else(|| "".to_string());
+
+    let last = array.next_back();
+
+    for value in array {
+        write!(sentence, ", {}", value).unwrap();
```
epage: > I'm just unsure if unwraping write! is the way to go. As far as I'm aware, it should be safe to write to a string but if I'm mistaken I can make it return an error instead.

I think you are correct.  The only failable operation should be a allocation (no IO to fail, formatting is compile-time) and allocations are panics.

I think I would prefer `expect`s though.  It documents your assumption to help people when auditing to consider whether we are right or not.

src/filters/array.rs
Goncalerta: I was about to say that they use no oxford comma, but then I went back and they do, in fact.

I don't know how that went over me, I really didn't notice it and I had looked at it at least a dozen of times.

epage: Very easy to miss., I obviously missed it the first time I reviewed the code and I was looking at tests which helps.

epage:
btw after this, if you are wanting a meatier thing to tackle, [re-doing the parser](https://github.com/cobalt-org/liquid-rust/issues/138) would be a good one.

There are two levels that could be done, I don't know which is right:
- A complete rewrite using a different parser
- Iterating on the current parser finding ways to make it easier and faster
  - faster: make `Token` not own strings, find ways to optimize the lexer, etc
  - easier: learn from `combine` or `nom` and refactor the parser to be combinators on top of the `Token`s.

Goncalerta:
Oh, sorry, its seems like I forgot to tell you I already made the changes you asked for.

As for the parser, I'm going to see how it is currently implemented and then look into `combine` and `nom` to see what can I do. Could you please tell me what are the features the parser needs to have that aren't already implemented, and other things to take into account?

Issue #145, opened 2017-11-08
Allow expressions in indexed variables


epage:
#141 added support for indexing literals `arr[0]` or `obj["name"]`.

This left out indexing by expressions like `result_string: {{ test_a[0][somevar] }}`

epage:
We'll need a new parser for this, see #138

epage:
Some of the API refactorings I'm working on for #95 might make this more attainable with the existing code base.

epage:
#214 implemented a subset of this, including taking care of the api-break.

PR #43, opened 2016-06-13
Make assign accept output expressions. (fixes #40)

johannhof:
Required some cleanups in `parser`, which prompted me to do some more
unrelated cleanups in other files as well.

Issue #40, opened 2016-06-08
Assign does not support piped expressions


bacoboy:
So I started working on an implmentation of `split` [here](https://github.com/bacoboy/liquid-rust/commit/3fdc801af325b9ca0cec64abd9f338c2181dac19).

The unit test passes just fine, but the resulting `Array` causes a problem downstream as you can see from the failed test:
```
---- split stdout ----
	thread 'split' panicked at 'called `Result::unwrap()` on an `Err` value: Lexer("\' is not a valid identifier")', src/libcore/result.rs:746
stack backtrace:
   1:        0x10bd959f8 - std::sys::backtrace::tracing::imp::write::hf68f1a220b61702c
   2:        0x10bd97c95 - std::panicking::default_hook::_$u7b$$u7b$closure$u7d$$u7d$::hb638acea7c29901b
   3:        0x10bd9781f - std::panicking::default_hook::h508c3dab3df347d6
   4:        0x10bd8c4d6 - std::sys_common::unwind::begin_unwind_inner::h17f9e42de6d55309
   5:        0x10bd8d28e - std::sys_common::unwind::begin_unwind_fmt::h039d18bd8498e1d0
   6:        0x10bd94cc7 - rust_begin_unwind
   7:        0x10bdbe510 - core::panicking::panic_fmt::h813eaa27a5810609
   8:        0x10bbc1ee8 - core::result::unwrap_failed::he9d1212db5110052
   9:        0x10bbc1828 - _<std..result..Result<T, E>>::unwrap::h7493e55a0acc9844
  10:        0x10bbcafcb - filters::split::h885acea6ebf1f10e
  11:        0x10bbe82cb - _<F as std..boxed..FnBox<A>>::call_box::h48dd10267ee7767d
  12:        0x10bbea79d - std::sys_common::unwind::try::try_fn::hc72e80a3e6bb322c
  13:        0x10bd94c5b - __rust_try
  14:        0x10bd94be3 - std::sys_common::unwind::inner_try::h74a189ca1fbe8e07
  15:        0x10bbeab84 - _<F as std..boxed..FnBox<A>>::call_box::h8f9b07ad31279db5
  16:        0x10bd96ee8 - std::sys::thread::Thread::new::thread_start::h40a33956f4cb7596
  17:     0x7fff9be2899c - _pthread_body
  18:     0x7fff9be28919 - _pthread_start
```
[...]

PR #221, opened 2018-11-05
Pest Parser

Goncalerta:
For the related issue, see #138 
For the unsquashed commit history see [parser-refactor-unsquashed](https://github.com/Goncalerta/liquid-rust/tree/parser-refactor-unsquashed).

I've been looking at some rust parsing libraries and I found `Pest` to seem really simple to use. So I decided to try it out and do some experiments, resulting in this pull request. 

This is just a draft, however, which means that there is no error handling yet, most tags were disabled (I changed the signature of `ParseTag` and `ParseBlock`), there are a lot of TODO's, issues and other things to improve in the code. 

Something I liked about pest is that most of the  error handling will be isolated from the rest of the parsing code. Also, Pest errors give information about line and column, which is a great help for providing more context to errors.

Implementing tags and blocks is now much easier, as the tokens are now "high level" (as a consequence of changing the signature). I made structs (`TagToken` and `TagBlock`) that replace `&[Token]` and `&[Element]` of the old parser, respectively. Instead of working with "low level" tokens, such as `Token::Pipe` or `Token::OpenSquare`, TagToken works with more ready to use ones, such as `Range` or `FilterChain`. TagBlock allows to easily parse everything inside a block, or to just get its raw content.

About performance, from what I've read, `Pest` is slower than `nom` (which is a natural tradeoff of its easiness of use). Still, I hope it's fast enough for this crate's purposes.

@epage Do you like this idea? Should this go forward or is it better to stick to regex/use something like `nom`?

## Features

- Rewrite the whole parser using [Pest](https://pest.rs/);

## Issues solved

- #138 
- #145 
- (Partially) #232
- #226 
- #227
- #242

## Breaking changes

Behavior
- Expressions no longer support tags (they weren't supposed to)
- More strictness in tokens accepted (Tags will raise an error if given a surplus of arguments. This is to alert the user for possible mistakes);

API
- Changed signature for tags and blocks.
- `compiler::parse` takes a`&str` directly as an argument, instead of requiring the midstep of `tokenize`.


epage:
Good work on this and good idea to do a vertical slice of this to judge feasibility and to get feedback.

Looking at it from the prospect of our goals
- Performance
  - Doesn't seem to do as many passes over the file as we currently do
  - Doesn't have extra allocations in the Token before moving to Renderables (I assume, its all internal to Pest)
- Make it easier to implement new features: Sure looks like it
- Easier to maintain: yup
- Concise top-level that reduces churn
  - This is my biggest area of concern
  - Is Pest in our public API? I haven't dug too much into these `TagToken` and `TagBlock`s
    - To be clear, I'm fine with having parsing utilitiies that live in a sub-crate that we can break compatibility on
    - I just don't want parser changes to break compatibility on Tags / Blocks and the overall top-level API
  - Can a third party add new tags / blocks without requiring updating our parser?
- File location information: Sounds like you've got it!
  - Is this provided in a way that we can update our filters, tags, and block implementations to track their location so semantic errors (and not just syntactic) can provide it?

I'll do a scan through and comment on details as well.

liquid-compiler/src/parser.rs
```diff
@@ -3,622 +3,411 @@
...
+    variable.extend(indexes);
+    variable
+
+    /* Why does this code appear on parse_expression() from the old parser?
+     * ```
+     * Some(&Token::Identifier(ref x)) if options.tags.contains_key(x.as_str()) => {
+     *      options.tags[x.as_str()].parse(x, &tokens[1..], options)
+     * }
+     * ```
+     * Are {{ expressions }} supposed to be treated as {% tags %} just because they have a tag name?
```
epage: - Issue: https://github.com/cobalt-org/liquid-rust/issues/40
- PR: https://github.com/cobalt-org/liquid-rust/pull/43
- Commit: https://github.com/cobalt-org/liquid-rust/commit/d2166407b6af0646169a7ade8deee6a9b7489246

I also dug into the Ruby code for liquid; I didn't do a good job of identifying if this behavior is in it.

My overall suspicion is that when filters were added to `assign`, the code was reused but in a way that accidentally made us also support tags for `{{}}` expressions.

We should note a breaking change for removing this.

liquid-compiler/src/grammar.pest
```diff
@@ -0,0 +1,56 @@
...
+TagStart = _{ (LINE_OR_SPACE* ~ "{%-") | "{%" }
+TagEnd =   _{ ("-%}" ~ LINE_OR_SPACE*) | "%}" }
+ExpressionStart = _{ (LINE_OR_SPACE* ~ "{{-") | "{{" }
+ExpressionEnd =   _{ ("-}}" ~ LINE_OR_SPACE*) | "}}" }
+
+Tag = !{ TagStart ~ Identifier ~ TagToken* ~ TagEnd }
+Expression = !{ ExpressionStart ~ FilterChain ~ ExpressionEnd }
+// Not allowing Tag/Expression Start/End might become a problem 
+// for {% raw %} tag: unclosed delimiters won't be accepted
+Raw = @{ (!(TagStart | ExpressionStart | TagEnd | ExpressionEnd) ~ ANY)+ }
```
epage: The downside to pest is trying to decipher it.
- I get that `~` is "followed by".
- I'm assuming `_` means non-addressable. 
- `@` I assume means addressable
- `!{`  from the docs seems to be never addressable.?

liquid-compiler/src/grammar.pest
```diff
@@ -0,0 +1,56 @@
...
+
+// Inner parsing        
+Identifier = @{ (ASCII_ALPHA | "_")+ }
+
+Variable = { Identifier 
+           ~ ( ("." ~ Identifier)
+             | ("[" ~ Value ~ "]")
+             )* 
+           }
+Value = { Literal | Variable }
```
epage: 👍  Looks like the grammar is being extended for #145

liquid-compiler/src/grammar.pest
```diff
@@ -0,0 +1,56 @@
...
+Filter = { Identifier ~ (":" ~ Value ~ ("," ~ Value)*)? }
```
epage: This is where we'd extend the grammar for #92 

Depending on architecture, we *might* be able to share named arguments with `for` blocks and others.

liquid-compiler/src/grammar.pest
```diff
@@ -0,0 +1,56 @@
...
+
+// Element-level parsing
+Element = _{ Expression | Tag | Raw } 
+
+TagStart = _{ (LINE_OR_SPACE* ~ "{%-") | "{%" }
+TagEnd =   _{ ("-%}" ~ LINE_OR_SPACE*) | "%}" }
+ExpressionStart = _{ (LINE_OR_SPACE* ~ "{{-") | "{{" }
+ExpressionEnd =   _{ ("-}}" ~ LINE_OR_SPACE*) | "}}" }
+
+Tag = !{ TagStart ~ Identifier ~ TagToken* ~ TagEnd }
```
epage: Without a pluggable parser of some sort, we end up requiring a catch-all `TagToken` that knows about every tags possible set of tokens and then we manually process those tokens later.

liquid-compiler/src/parser.rs
```diff
@@ -3,622 +3,411 @@
...
-const NOTHING: Option<&str> = None;
+use pest::Parser;
 
-pub fn unexpected_token_error<S: ToString>(expected: &str, actual: Option<S>) -> Error {
-    let actual = actual.map(|x| x.to_string());
-    unexpected_token_error_string(expected, actual)
-}
+#[derive(Parser)]
+#[grammar = "grammar.pest"]
+pub struct LiquidParser;
```
epage: This is an area that hurts maintainability.  Macros like this make it hard to tell what all is available in the API.

liquid-compiler/src/grammar.pest
```diff
@@ -0,0 +1,56 @@
...
+LINE_OR_SPACE = _{ NEWLINE | WHITESPACE }
+LiquidFile = ${ SOI ~ Element* ~ EOI }
+
+// Element-level parsing
+Element = _{ Expression | Tag | Raw } 
+
+TagStart = _{ (LINE_OR_SPACE* ~ "{%-") | "{%" }
+TagEnd =   _{ ("-%}" ~ LINE_OR_SPACE*) | "%}" }
+ExpressionStart = _{ (LINE_OR_SPACE* ~ "{{-") | "{{" }
+ExpressionEnd =   _{ ("-}}" ~ LINE_OR_SPACE*) | "}}" }
```
epage: I didn't notice in the Ruby liquid, does it support `{{-` and friends for stripping whitespace?

epage:
Not finding it but I thought I remember seeing a comment about preserving `{{ value }}` being treated differently than `{{ value | filter }}`.  I don't know if I'd worry about that myself.

epage:
I think I'm done with my high level pass.  Please feel free to express your thoughts; don't take me being unsure about the compatibility aspects as a go/no-go on this.

epage:
RE Keeping our API stable

Looking at pest, it seems reasonable for us to be able t find a way to keep it out of our API and allow changing parsing engines without a problem.  

The big question is whether to try to do all of the changes in a single PR or for us to do it over time.  A single PR means we touch all of the tags / blocks in one go but I also find its harder to design and iterate when doing that.  Probably better to do multiple passes.

So after reading my comments, if you feel this is the best way moving forward, go for it! I have nothing against it.

liquid-compiler/src/parser.rs
epage: At my quick glance at the Pest docs, is Pest all within a trait and this grammar is used for implementing the trait?  So its not like we actually have a bunch of hidden API calls that are created.  Well, I guess the Rules would be.

liquid-compiler/src/parser.rs
```diff
@@ -3,622 +3,411 @@
...
-                Token::Identifier(ref name) if options.blocks.contains_key(name.as_str()) => {
-                    stack.push("end".to_owned() + name);
-                }
+) -> Result<Box<Renderable>> {
+        let mut nesting_depth = 1;
+        let mut block = TagBlock::new();
 
-                Token::Identifier(ref name) if Some(name) == stack.last() => {
-                    stack.pop();
+        while let Some(element) = next_elements.next() {
```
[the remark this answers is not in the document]
epage: To be clear, I feel like we could implement this nom-like approach in Pest.   As long as it gives us where the parser left off, we could get this going.  We'd have to pass along that information as well as line / column information since pest wouldn't realize we were picking up again within the same file.

liquid-compiler/src/grammar.pest
```diff
@@ -0,0 +1,56 @@
...
+// Literals 
+StringLiteral = @{ ("'" ~ (!"'" ~ ANY)* ~ "'")
+				 | ("\"" ~ (!"\"" ~ ANY)* ~ "\"") }
+                 
+IntegerLiteral = @{ ASCII_DIGIT+ }
+FloatLiteral = @{ ASCII_DIGIT+ ~ "." ~ ASCII_DIGIT+ }
+
+BooleanLiteral = @{ "true" | "false" }
+
+Literal = { StringLiteral | IntegerLiteral | FloatLiteral | BooleanLiteral }
```
epage: Looks like `nil` and `null` are supported as literals
https://github.com/Shopify/liquid/blob/master/lib/liquid/expression.rb

liquid-compiler/src/grammar.pest
epage: Looks like a `Value` can also be a range?

https://github.com/Shopify/liquid/blob/master/lib/liquid/expression.rb#L28
https://github.com/Shopify/liquid-c/blob/master/ext/liquid_c/parser.c#L131

liquid-compiler/src/grammar.pest
epage: Apparently it is
https://shopify.github.io/liquid/basics/whitespace/

liquid-compiler/src/grammar.pest
```diff
@@ -0,0 +1,56 @@
...
+           //| Literal 
+           //| (Variable ~ !"|") // If Variable is followed by "|", then it is a filterchain, and thus must not be consumed.
+           // On the other hand, another possibility would be to make filterchain's filters mandatory, thus any filterchain without filters degenerates to filter
+           //| FilterChain 
+           //}
+
+Range = { "(" ~ Value ~ ".." ~ Value ~ ")" }
+ComparisonOperator = { "==" | "!=" | "<>" | ">" | "<" | ">=" | "<=" | "contains" }
+LogicalOperator = { "and" | "or" }
+AssignmentOperator = { "=" }
```
epage: Another token is `?`

The problem is I do not see a single place where [we](https://github.com/cobalt-org/liquid-rust/search?utf8=%E2%9C%93&q=Question&type=), [the ruby implementation](https://github.com/Shopify/liquid/search?q=%3Aquestion&unscoped_q=%3Aquestion), or the [c implementation](https://github.com/Shopify/liquid-c/search?q=TOKEN_QUESTION&unscoped_q=TOKEN_QUESTION) use it.

So probably just hold off?

Goncalerta:
>   * Is Pest in our public API? 

I don't think so, at least not after this pull becomes complete. The parser's public API is going to be `parse(text: &str, options: &LiquidOptions) -> Result<Vec<Box<Renderable>>>`, which does not leak anything about the parser itself, `TagBlock` and `TagTokens`, which were created exactly to encapsulate Pest's `Pairs`, providing useful methods to easily implement new tags/blocks.

>     * I just don't want parser changes to break compatibility on Tags / Blocks and the overall top-level API

Once this PR is done (which will inevitably break the compatibility on Tags/Blocks) I hope no more big compatibility breaks will occur. That could only happen, I think, if `TagBlock` and `TagTokens` needed to change their API and I hope they are stabilized once this is finished. 

>   * Can a third party add new tags / blocks without requiring updating our parser?

Well, right now, this draft is very limited. But my goal is to make `TagBlock` and `TagTokens` flexible enough to allow this.

> * File location information: Sounds like you've got it!
>   
>   * Is this provided in a way that we can update our filters, tags, and block implementations to track their location so semantic errors (and not just syntactic) can provide it?

I still have to think about it, as I haven't put much thought in the error handling yet. Semantic errors would be to, for example, try to access a nonexisting variable, right? That can only be known during rendering, as it depends on the context.

As such, the only way I could see this happening is to update structs like `Variable` to store their location. Then, when creating `Variable`s, the parser would pass in their location. Once an error happens, the error crate could look into that information and produce an error message.

liquid-compiler/src/grammar.pest
Goncalerta: - `~` is followed by.
- `_` means that it doesn't produce a `Pair`. So when 'unwrapping' (as in, calling into_inner() ) `Tag`, for example, I would get an iterator with `Identifier` followed by `TagToken`s.
- `@` means the rule is 'atomic' ( as in, it can't be subdivided into other rules). This means that instead of Raw being an iterator over `ANY` (means any character is accepted), it is the text as a whole. It won't be subdivided on an iterator over characters.
- `!something` means to look into the next token and make sure its not `something`. So Raw will be `ANY`thing as long as `ANY` is not, for example, part of a `{{`.

liquid-compiler/src/parser.rs
Goncalerta: Hmm, I've used `pub struct` but now that I look at it, it seems its not necessary to be `pub`.

liquid-compiler/src/grammar.pest
Goncalerta: I'm not sure about Ruby liquid, but it does indeed happen in the old parser of liquid-rust and also in [liquidjs](https://www.npmjs.com/package/liquidjs). I think there is no need to remove that, as it may be helpful.

Goncalerta: Oh it seems you already found that out. I didn't refresh the page, so I didn't see the new messages.

liquid-compiler/src/grammar.pest
Goncalerta: > Not finding it but I thought I remember seeing a comment about preserving `{{ value }}` being treated differently than `{{ value | filter }}`. I don't know if I'd worry about that myself.

Probably you're refering to the `grammar.pest` comments. That was a suggestion I made in an earlier phase so `Value` and `Filterchain` would be parsed differently on TagToken, so one wouldn't "eat" the other when parsing. But right now `Filterchain` "eats" everything and `TagTokens` takes care of "unwrapping" the value out of the filterchain so I think there won't be a problem anymore.

Goncalerta:
> Not finding it but I thought I remember seeing a comment about preserving `{{ value }}` being treated differently than `{{ value | filter }}`. I don't know if I'd worry about that myself.

Probably you're refering to the `grammar.pest` comments. That was a suggestion I made in an earlier phase so `Value` and `Filterchain` would be parsed differently on TagToken, so one wouldn't "eat" the other when parsing. But right now `Filterchain` "eats" everything and `TagTokens` takes care of "unwrapping" the value out of the filterchain so I think there won't be a problem anymore.

liquid-compiler/src/parser.rs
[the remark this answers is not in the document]
Goncalerta: I'm thinking that maybe `TagBlock` could be wrapping the iterator of the whole "stream of elements" and its functions (that would be used in block tags creation) would silently check if it is parsing the close tag, stopping at that point.

liquid-compiler/src/grammar.pest
epage: > !something means to look into the next token and make sure its not something. So Raw will be ANYthing as long as ANY is not, for example, part of a {{.

I chose a bad example.  That one was obvious.  The confusion comes in when `!` is reused in other syntax like

> Tag = !{ TagStart ~ Identifier ~ TagToken* ~ TagEnd }

epage:
> Well, right now, this draft is very limited. But my goal is to make TagBlock and TagTokens flexible enough to allow this.

I'm looking forward to your thoughts on how this will be possible.  Like I said in a follow up, I'm not bothered if it doesn't work out because I think we'll still have a path forward for it.

liquid-compiler/src/grammar.pest
Goncalerta: Oh. Basicaly, when I do `LiquidFile = ${ SOI ~ Element* ~ EOI }`, the $ prevents implicit whitespace. This is needed because otherwise all spaces after closing an Expression/Tag would be trimmed istead of being parsed as Raw. But $ will disable whitespace in the inner rules too (in this case the whole file because we're wrapping `LiquidFile`  rule with the $).

The !, on the other hand, is the exception to `$`: no matter whether the outter rule has `$` or not, if the inner rule is marked with ! it will never prevent implicit whitespace. This is needed inside Tags and Expressions so the whitespaces are trimmed as expected.

epage:
> What else must be taken into account to achieve the desired flexibility?

What if all tags / block knowledge is third party, how dwell can our parser handle that?

Goncalerta:
> > What else must be taken into account to achieve the desired flexibility?
> 
> What if all tags / block knowledge is third party, how dwell can our parser handle that?

In that extreme case I guess it could have a function like `parse_next_element_as_str` which would yeld the next element without parsing it (it would return, for example, "{% tagname something blahblahblah %}"). And the user could make the parsing on his own.

TagTokens could similarly have a `to_str` that would yeld whatever wasn't parsed yet as a &str.

This is the only way I'm seeing if we can't really do any assumptions at all regarding how it would be parsed. But this would also mean the user would be responsible for keeping track of line and column numbers in case they want to raise a semantic error during their custom parse.

epage:
> In that extreme case I guess it could have a function like parse_next_element_as_str which would yeld the next element without parsing it (it would return, for example, "{% tagname something blahblahblah %}"). And the user could make the parsing on his own.

So again, I don't think we need to solve this in the first iteration but I want us to be able to eventually support new, unthought of tags/blocks without having to modify liquid-rs.

When we get there, another approach is we remove SOI / EOI and parse up until the next tag / block.  We then hand over where we left off in the `str` to the tag/block (plus line/column info).  Once it reaches the end of the block, it returns where it left off in the str and we start parsing again.

liquid-compiler/src/grammar.pest
Goncalerta: I'm gonna mark this as resolved as there is an issue opened for it.

liquid-compiler/src/parser.rs
```diff
@@ -198,85 +188,84 @@ fn parse_tag(
...
+            renderables.push(parse_element(element, self.iter, options)?);
         }
         Ok(renderables)
     }
-    
+
     // Probably a to_str might be implemented by checking the start of 1st pair and end of the last.
     // Then just create a slice directly from the input text.
-    pub fn to_string(&self) -> String {
+    pub fn to_string(&mut self) -> Result<String> {
```
epage: Should this just take ownership at this point?

liquid-compiler/src/parser.rs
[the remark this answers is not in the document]
Goncalerta: I made some changes to the `TagBlock`. Now, the check for the end tag occurs inside of its `next` function (in other words, when a new element is being requested from the "stream of elements"). At the end, the `TagBlock` closes, leaving the iterator at the right place for the next element to be parsed. 

Is this the type of approach desired?

liquid-compiler/src/parser.rs
Goncalerta: The problem with taking ownership of `TagBlock` is that it would be unavailable to call `close()` at the end of `parse_tag()`. So the `close()` function is the one that currently can take ownership.

Unless we find a better approach than the `close()` function to ensure the block is parsed to its end, I don't think we can change this.

Goncalerta: Regarding the first thought: the TagBlock never exposes Pest's API (and thus `next()`). As such, there is no way for `raw` to do this, as the only function that would be available besides `to_string` is `parse`, which would actually parse the elements instead of returning them as a string. There may also be other tags (eventually) that use this function (or something similar) if they need something different than just directly parsing the elements of the block.

On the second thought: that might be a possibility, although I'm not really sure if there is a way to do the error check without still needing something like the `close()` function (and thus a `&mut`). Do you have any ideas?

liquid-compiler/src/parser.rs
```diff
@@ -3,622 +3,408 @@
...
-        let mut post = Object::new();
-        post.insert("number".into(), Value::scalar(42i32));
-        context.stack_mut().set_global("post", Value::Object(post));
-
-        let result = parse_expression(&tokens, &null_options()).unwrap();
-        let result = result.render(&mut context).unwrap();
-        assert_eq!("42", result);
+    fn close(mut self) -> Result<()> {
+        while self.nesting_depth != 0 {
+            self.next()?;
```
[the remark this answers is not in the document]
Goncalerta: Good point. I've made the change, but now I'm unsure if the way I did is the idiomatic way to go in rust. Is it fine to make `while` loops with an empty body or is there a better approach?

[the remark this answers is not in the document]
epage: I thought there would be a better way also.  I looked for `drain` but it appears we don't have that for `Iter`, only `Vec`.

liquid-compiler/src/parser.rs
```diff
@@ -3,622 +3,406 @@
...
-}
+#[derive(Parser)]
+#[grammar = "grammar.pest"]
+struct LiquidParser;
 
-pub fn unexpected_token_error_string(expected: &str, actual: Option<String>) -> Error {
-    let actual = actual.unwrap_or_else(|| "nothing".to_owned());
-    Error::with_msg(format!("Expected {}, found `{}`", expected, actual))
-}
+type Pair<'a> = ::pest::iterators::Pair<'a, Rule>;
```
epage: Sorry, I've not had a chance to read up on pest's design.  What is a `Rule`?  Are we exposing pest in our API via this?

Either way, I think in the first pass, I'd be fine with it and we can experiment and iterate on designs that slowly work to remove pest from the public API.

Goncalerta: In the first place, this is not a `pub type`, so there is no need to worry about being exposing pest in the API. I just use that type to make function declarations shorter.

`Rule`, if I understand it correctly, is the enum of rules generated by the macro `#[grammar = "grammar.pest"]`, such as `Rule::Expression` or `Rule::Literal` (there is one for each rule defined in `grammar.pest`).

Because `Rule` is only generated in our code, it is bound to the type `pest::iterators::Pair` by a generic type.

Goncalerta: Maybe I didn't understand what you were asking. Is the `Rule` enum itself on the public API?

Good question, I assumed it wasn't up to this point, but I guess if it is we'll have to isolate it.

Goncalerta: Macros `#[derive(Parser)]` and `#[grammar = "grammar.pest"]` are now isolated in a module `pest`, so they don't put something unexpected on the public API.

liquid-compiler/src/grammar.pest
epage: Created https://github.com/cobalt-org/liquid-rust/issues/224 for ranges as values

liquid-compiler/src/parser.rs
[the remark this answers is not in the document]
epage: We'll see how it works.

liquid-compiler/src/grammar.pest
epage: Created https://github.com/cobalt-org/liquid-rust/issues/225

liquid-compiler/src/parser.rs
epage: So `struct LiquidParser` isn't public, `type Pair` isn't public, and `Rule` (hidden in macro) is not public.

On the other hand, `TagToken` and `TagBlocks` are public and return `Pair` which I assume also returns `Rule`.  That would make them public in every way except being able to refer to their type (which they can refer to `Pair` through pest's API).  Until this is resolved, we should just make them `pub` so people can reference their type.


Weird, Travis isn't reporting any warnings for `private-in-public` which is `warn` by default and out Travis config elevates to `deny`.

Goncalerta: `TagToken` and `TagBlock`  are public, but they never return `Pair` in their public API, nor I intend them to. I want to keep private every function that would ever return something that exposes `Pest`. The functions that return `Pair` are supposed to be internal helpers only.

epage: Sorry, I missed the lack of `pub` on those functions.

Honestly, when implementing the rest of the tags and blocks, I can't see how we could implement them without either these functions being `pub` or effectively implementing every tag and block's parse logic inside of `TagBlock` and `TagTokens`.

epage: Again, to be clear, any of this is better than what we have now and we ,an always iterate on it later.

Thinking about that, id go as far as recommending we have pest in the api for now rather than blocking this on trying to find ways to ensure it isn't. Itll give us the benefits of this change sooner and allow us to better see how removing pest impacts the design.

liquid-compiler/src/parser.rs
```diff
@@ -3,622 +3,420 @@
...
-            .stack_mut()
-            .set_global("foo", Value::scalar("number"));
-
-        let result = parse_expression(&tokens, &null_options()).unwrap();
-        let result = result.render(&mut context).unwrap();
-        assert_eq!("42", result);
+    pub fn to_string(&mut self) -> Result<String> {
+        let mut s = String::new();
+        while let Some(element) = self.next()? {
+            s.push_str(element.as_str());
```
epage: This is used by `raw`, right?

What if we have
```liquid
{raw}
{raw}
{endraw}
```

Won't the use of `next` cause this to fail?

Goncalerta: Yep. On the current parser, though, the problem is even bigger.
Not only this also happens (tries to nest raw tags) but it doesn't even require to close the block.
The output is like this:
`{%raw%}{%endraw%}`
Which is worse because not only it nests them, but it also allows the blocks to be freely unclosed.

I didn't bother much with this case because `{%endraw%}` will already be unable to be used normally inside a `raw` block no matter what. I assumed writting `{%raw%}` inside `raw` blocks also wouldn't be needed in a practical scenario. If you prefer, I can try to create a way for blocks to block nesting.

epage: If it is an existing problem, I recommend creating an issue and moving on.

epage: Thanks!

liquid-compiler/src/parser.rs
Goncalerta: To be honest, I think bringing Pest to the API might do more harm than good.

  - Users would need to understand Pest and the underlying structure of the parser in order to understand how to create tags.
  - Bringing `Rule` to the API seems like a bad idea and hard to use because of it being generated by a macro.
  - For `TagTokens`, because many tokens end up being encompassed by `Filterchain`, users would need to manually do the repetitive task of decomposing that token to the one they want. For example, if they desired to expect a `Value`, they would need to access the inner `Pair`s in the `Pair` they've got (the one with `Filterchain` as a rule) and manually test if the input hadn't any unexpected filters, manually returning an error if they had.
  - I suppose it would be harder to implement error handling without requiring the user to constantly manually use functions in their tags that they could easily forget.
  - Removing Pest from the API would imply yet another breaking change.

Given this, I will try to find another way, even if that is simply encapsulating `Pair` itself instead of the iterator over pairs. I suppose that would keep all the flexibility from `Pair` (and maybe even add some utility functions) without exposing the user to something so 'internal' to the parser.

epage: > Users would need to understand Pest and the underlying structure of the parser in order to understand how to create tags.

Only tag authors and, as I said, this is a short-term thing.  My priority is to get the benefits of this PR.  We can then experiment on how to refine it from there.

> Removing Pest from the API would imply yet another breaking change.

Yes, we have further breaking changes down the line.  I'm working on some additional ones.  I'd prefer we iterate towards a smaller surface of breaking changes rather than trying to do it all at once.

If nothing else is pushing a release, we could just wait on a release until one or two of these further PRs is in, consolidating these breaking changes.

> Given this, I will try to find another way, even if that is simply encapsulating Pair itself instead of the iterator over pairs. I suppose that would keep all the flexibility from Pair (and maybe even add some utility functions) without exposing the user to something so 'internal' to the parser.

As I said, I'd recommend on getting something in and then having smaller, more targeted PRs to refine how we do this.  Smaller PRs are easier to follow the conversation for, faster to get in, generally less depressing for contributors, etc. 

If you have an idea that is reasonable to execute on within this PR, go for it.  I just recommend we don't try to solve all problems in this PR.

Goncalerta: I tried to implement all tags with the new API from this PR to see if it was possible to do so without the use of Pairs.

I got stuck, though, as I'm having trouble implementing tags that formerly (in the old parser) depended on `split_block`. I don't think that even after exposing `Pair` in the API I would be able to solve this problem. Do you have any ideas on this?

src/tags/include_tag.rs
```diff
@@ -29,25 +30,28 @@ impl Renderable for Include {
...
 ) -> Result<Box<Renderable>> {
-    let mut args = arguments.iter();
-
-    let name = match args.next() {
-        Some(&Token::StringLiteral(ref name)) => name,
-        Some(&Token::Identifier(ref s)) => s,
-        arg => return Err(unexpected_token_error("string", arg)),
+    let name = arguments.next().unwrap_or_else(|| panic!("Errors not implemented. Token expected."));
+    // TODO: make `name` a &str instead
+    let name = if let Ok(name) = name.expect_identifier() {
```
epage: Just to remind us of this todo

src/tags/include_tag.rs
```diff
@@ -29,25 +30,28 @@ impl Renderable for Include {
...
+        name.to_string()
+    } else if let Ok(name) = name.expect_literal() {
+        // This will allow non string literals such as 0 to be parsed as such.
+        // Is this ok or should more specific functions be created?
```
epage: Not dug into this yet but don't we need a way to turn a literal into a `Scalar` which requires knowing if it is a string literal or not?  Can't we leverage that?

liquid-compiler/src/parser.rs
```diff
@@ -3,622 +3,445 @@
...
-        )),
+    let mut tag = tag.into_inner();
+    let name = tag
+        .next()
+        .expect("A tag starts with an identifier.")
+        .as_str();
+    let mut tokens = tag.map(TagToken::from);
+
+    if options.tags.contains_key(name) {
+        options.tags[name].parse(name, &mut tokens, options)
```
epage: Why is tokens passed as a `&mut` rather than transferring ownership?

liquid-compiler/src/parser.rs
epage: The main one I have is to invert the control flow.  Instead of a block relying on the parser to handle block content and find end of block, we pass the iterator to the block.

In the case of an `if` block, it can iterate until the next tag
- if its `else` or `elsif`, change state and continue.
- if it is an `endif`, then stop processing and give control back to the caller.
- if it is a start tag, delegate the iterator to that block for processing, resuming where it left off when it finishes

This also has the benefit of automatically handling nesting for us.  Instead of tracking nesting depth, we use the call stack.  And `if` in an `if` will consume the iterator until it finds an `endif`, hand control back, and then that `if` will resume from there until it gets to an `endif`.

This does require us to trust the tags/blocks to be written correctly.  Utilities can be provided to help share the termination logic so we don't have to duplicate it, like a `consume_until`.  We might want different styles for different cases rather than a single helper trying to handle all styles of blocks.

As we refactor towards a stable API, these helpers would be moved off into a sub-crate that is free to break compatibility without impacting the use of the top level `liquid` crate.

epage:
Again, thanks for looking into this.  There are a lot of tricky or annoying cases this requires and you are doing a great job staying on top of it all.

liquid-compiler/src/parser.rs
Goncalerta: If I remember well, the function signature accepts an `Iterator<Item=TagToken>` as the parameter, which is a trait. This meant that in order to pass a concrete object, I would need instead to pass a pointer (in this case, mutable reference) to it. This probably may be fixed by changing the signature to generics.

epage: Ah, good point.

Tags/blocks can't accept generics because they need to be Trait Objects.

One alternative is to pass in a slice instead.  Not sure which would be better.

liquid-compiler/src/parser.rs
epage: Also, I am fine with you reverting it back to something more akin to the original parser where we `collect` all the elements for a given block which allows the block to do a split.  We could then refactor to my proposal or another idea at a later point.

liquid-compiler/src/parser.rs
Goncalerta: Passing a slice would require to collect the iterator and most tags would probably just turn the slice into an iterator again.

I guess having iterators would be the best option, at least for now.

Goncalerta: Besides, slices are still tecnically just references so ownership is not transferred.

epage: So maybe I should step back and explain what felt weird to me with the API.  To be clear, this is to express my reasoning  and not to say which way it has to be done.

> Besides, slices are still tecnically just references so ownership is not transferred.

A `&mut` is ambiguous about where in the iterator you are starting and where you are ending ie what is the expectation for how each side of the API works.

A `&` slice would better convey "everything in this is for you" (unless we did other stuff like having the return value include the slice).

src/tags/include_tag.rs
Goncalerta: The thing is currently there is a function `parse_literal` that automatically works for any literal, `String`, `Boolean`, `Float`, `Integer`, turning them into the respective `Scalar`.

The function `expect_literal` allows any literal to be parsed, returning a `Value`. I think this is fine for most use cases.
Now for cases like this one, I'm unsure what the behavior of liquid is supposed to be. If its fine for it to accept other literals as well, this is the way to go. Otherwise, more functions must be created, to expect specific types of Literals.

liquid-compiler/src/parser.rs
Goncalerta: What about using a `Box`? Would that better express the meaning of the API?

Goncalerta: Aparently that raises problems because `Box` requires a `'static` lifetime.

Goncalerta: I keep getting this error message
```
error[E0495]: cannot infer an appropriate lifetime for lifetime parameter `'i` due to conflicting requirements
   --> liquid-compiler\src\parser.rs:187:23
    |
187 |     let mut tag = tag.into_inner();
    |                       ^^^^^^^^^^
    |
note: first, the lifetime cannot outlive the lifetime 'a as defined on the function body at 178:14...
   --> liquid-compiler\src\parser.rs:178:14
    |
178 | fn parse_tag<'a>(
    |              ^^
    = note: ...so that the types are compatible:
            expected pest::iterators::Pair<'_, _>
               found pest::iterators::Pair<'a, _>
    = note: but, the lifetime must be valid for the static lifetime...
    = note: ...so that the expression is assignable:
            expected std::boxed::Box<(dyn std::iter::Iterator<Item=parser::TagToken<'_>> + 'static)>
               found std::boxed::Box<dyn std::iter::Iterator<Item=parser::TagToken<'_>>>
```

epage: I think I reproduced your problem on the playground and after doing some searching, I found the part of lifetimes I frequently forget

`Box<Iterator<Item=&'a i32>>` is implicitly `Box<Iterator<Item=&'a i32> + 'static>`, requiring the type you are boxing to be `'static`.  Since your iterator is not, you need to mention the lifetime that the iterator is valid for: `Box<Iterator<Item=&'a i32> + 'a>`

See https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=37b360c69c4e60658d3012309a1b341e

epage: The place I found the answer: https://www.reddit.com/r/rust/comments/325u5a/right_way_to_return_an_iterator_over_a_slice/?st=johq7l62&sh=ea03b4f9

Goncalerta: Thanks! I had never seen that syntax `+ 'lifetime` before.
I tried to implement that, which led to using a lot of `<'a>` lifetimes in `tag.rs`. Still, there are errors I'm not being able to overcome.
```
error[E0477]: the type `tag::FnTagParser<'a>` does not fulfill the required lifetime
  --> liquid-compiler\src\tag.rs:54:10
   |
54 | impl<'a> ParseTag<'a> for FnTagParser<'a> {
   |          ^^^^^^^^^^^^
   |
   = note: type must satisfy the static lifetime

error[E0477]: the type `tag::BoxedTagParser<'a>` does not fulfill the required lifetime
  --> liquid-compiler\src\tag.rs:76:10
   |
76 | impl<'a> ParseTag<'a> for BoxedTagParser<'a> {
   |          ^^^^^^^^^^^^
   |
   = note: type must satisfy the static lifetime

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0477`.
```

I am very confused, as I don't understand why those would be required to satisfy the static lifetime.

Goncalerta: Threads, you say? That makes a lot of sense actually, as `ParseTag` implements `Send + Sync`. Maybe the best thing to do right now is to avoid all these changes and leave the iterator as a `&mut` or slice and think about boxing it in a future PR.

epage: I agree.  Better to get an imperfect solution in now and then iterate on it.

liquid-compiler/src/parser.rs
```diff
@@ -3,652 +3,562 @@
...
+    Raw(Raw<'a>),
+    Tag(Tag<'a>),
+    Expression(Exp<'a>),
+}
+impl<'a> From<Pair<'a>> for BlockElement<'a> {
+    fn from(element: Pair<'a>) -> Self {
+        match element.as_rule() {
+            Rule::Raw => BlockElement::Raw(element.into()),
+            Rule::Tag => BlockElement::Tag(element.into()),
+            Rule::Expression => BlockElement::Expression(element.into()),
```
epage: Doesn't `BlockElement` duplicate `parse_element`?

liquid-compiler/src/parser.rs
```diff
@@ -3,652 +3,562 @@
...
+        next_elements: &mut TagBlock,
+        options: &LiquidOptions,
+    ) -> Result<Box<Renderable>> {
+        let (name, tokens) = (self.name, &mut self.tokens);
+
+        let next_elements = &mut next_elements.iter;
+
+        if options.tags.contains_key(name) {
+            options.tags[name].parse(name, tokens, options)
+        } else if options.blocks.contains_key(name) {
```
epage: Doesn't `Tag` duplicate `parse_tag`?

liquid-compiler/src/parser.rs
epage: From my quick glance, it seems like it'll be a good starting point.  I'd say, go ahead and polish this and let's get it in.  At some point I'm going to need to pull down your branch to get a better look at it than what I can do within github.

liquid-compiler/src/parser.rs
Goncalerta: On the most part yes, and they can definitely be put together. I just hadn't thought very much on the best way to do so yet, as I was more focused on getting the if/for/case tags implementable.

I'm actually thinking about taking the idea of those structures such as `Tag` and "Expression" a bit further and do that for all 'unparsed' states of other elements of the file, replacing the current functions by associated functions to these 'unparsed' states. I think that would bring more consistency to the whole parser and also allow to make those structures public (if ever needed) without having to leak Pest's API, as well as more flexibility and type safety (because we would no longer work directly with `Pair`s, which could be anything, but with structures encapsulating them). But this is something for a future PR.

liquid-error/src/error.rs
```diff
@@ -94,6 +94,15 @@ pub struct Error {
     inner: Box<InnerError>,
 }
 
+impl<R> From<pest::error::Error<R>> for Error
```
epage: So you'll find we have a lack of `From`s.  This is because they become part of the public API and I'm trying to avoid implementation details in the public API.

epage:
btw opened #232  to help track some of the less obvious requirements we'll eventually need to handle for reporting line numbers in errors.  I don't think its needed in this PR but we should get to it soon after.

src/tags/if_block.rs
```diff
@@ -185,7 +185,40 @@ impl Renderable for Conditional {
     }
 }
 
-fn parse_atom_condition(arguments: &mut TagTokenIter) -> Result<Condition> {
+struct PeekableTagTokenIter<'a> {
```
Goncalerta: I don't really think this is the best approach. However, I wasn't able to find another way to have both `peek` and `expect_next` accessible to parse the condition.

src/tags/include_tag.rs
Goncalerta: Now that I look at this again, I realize that, at the end of the function, `name` would be casted into a `String` anyway. So there is no benefit in trying to keep it as a reference here.

liquid-compiler/src/parser.rs
Goncalerta: I'm marking this as resolved, as I think that if something is to change about the way `close()` works, it's better to do it in a future PR.

Regarding the other thought, it is already solved.

Goncalerta: Actually, as I was writting this I remembered about the `Drop` trait.
As it may be a small change, I will see if it can be used to fix this in this PR. Otherwise, I'll let this for the future.

Goncalerta: Ok, this is now solved and `TagBlock` is no longer passed to `block` functions as a `&mut`.

epage:
FYI I'm working on porting integration tests over from the ruby implementation to be able to act as a conformance test.  Once we get the improved parser in, there will be lots of opportunities to tweak it to behave more correctly :)

Goncalerta:
On `benches/handlebars-liquid.rs` line 9 we have:
```

**Comment by Goncalerta** [comment]
On `benches/handlebars-liquid.rs` line 9 we have:
```

Goncalerta:
On `benches/handlebars-liquid.rs` line 9 we have:
```
static ITERATE: &'static str = "<html>
  <head>
    <title>{{year}}</title>
  </head>
  <body>
    <h1>CSL {{year}}</h1>
    <ul>
    {% for team in teams %}
      <li class=\"{% if forloop.index0 == 0 %}champion{{% endif %}\">
      <b>{{team.name}}</b>: {{team.score}}
      </li>
    {% endfor %}
    </ul>
  </body>
</html>";
```

Notice that part `champion{{% endif %}` is that extra `{` before the tag supposed to be there? Is liquid supposed to accept that as a raw instead of raising an error? Because right now in this PR it doesn't and that's causing travis to fail.

Goncalerta:
Besides that, I think everything is ready for this pull request (after you've done your usual review of course).

Now, about squashing the commits, I feel it would be kinda weird to have all those changes in just a single commit. However, I also don't think there would be any meaningful way to split this PR in many commits, as the changes depend on each other (for example, refactoring the parser without refactoring the tags at the same time (or vice-versa) wouldn't make much sense). What do you think that should be done? Just squash them all into one?

epage:
> Notice that part champion{{% endif %}. Is that extra { before the tag supposed to be there? Is liquid supposed to accept that as a raw instead of raising an error? Because right now in this PR it doesn't and that's causing travis to fail.

That is a typo and I have a PR that fixes it.  On the other hand, that PR is also adding conformance tests and I think one of them does something similar.  I'd recommend waiting for #239 to be merged.

epage:
> Now, about squashing the commits, I feel it would be kinda weird to have all those changes in just a single commit. However, I also don't think there would be any meaningful way to split this PR in many commits, as the changes depend on each other (for example, refactoring the parser without refactoring the tags at the same time (or vice-versa) wouldn't make much sense). What do you think that should be done? Just squash them all into one?

If it is all parser work and dependent on each other, I'd just squash it.

I'll try to take a look at this later today.

Goncalerta:
The only thing that I think can be put in a different pull request is 76cdde4aadfd96aaa4e405af57c7807146b3433f, where a fixture is fixed.

epage:
Sounds like a good idea.

epage:
Performance:
- `parse_template` went from 42ms to 33ms
- `parse_text went from 592ns  to 1654ns
- `bench_parsing_basic_template` went from 31ms to 27ms
- `bench_big_loop_big_object` went from 866ms to 967ms

Source
- https://travis-ci.org/cobalt-org/liquid-rust/jobs/459806284
- https://travis-ci.org/cobalt-org/liquid-rust/jobs/459831175

Caveat: this is relying on Travis which is not very predictable in its performance.

src/tags/capture_block.rs
```diff
@@ -40,25 +40,29 @@ impl Renderable for Capture {
...
-    let t =
-        Template::new(parse(tokens, options).trace_with(|| format!("{{% capture {} %}}", &id))?);
+    let template = Template::new(
+        tokens
+            .parse_all(options)
+            .trace_with(|| format!("{{% capture {} %}}", &id))?,
+    );
 
-    Ok(Box::new(Capture { id, template: t }))
+    tokens.assert_empty();
```
epage: Unsure if I missed this, is this in every block?  If not, why is this different?

src/tags/capture_block.rs
```diff
@@ -40,25 +40,29 @@ impl Renderable for Capture {
...
-        return Err(unexpected_token_error("`%}`", t));
-    };
+    let id = arguments
+        .expect_next("Identifier expected")?
+        .expect_identifier()
+        .map_err(TagToken::raise_error)?
+        .to_string();
+
+    // no more arguments should be supplied, trying to supply them is an error
+    if let Some(token) = arguments.next() {
```
epage: Unsure if I missed this, is this in every tag/block?  If not, why is this different?

liquid-compiler/src/parser.rs
```diff
@@ -3,477 +3,831 @@
...
-                Some(&&Token::Pipe) | None => break,
-                _ => {
-                    return Err(unexpected_token_error(
-                        "`,` | `|`",
-                        Some(iter.next().unwrap()),
-                    ));
-                }
+        if element.as_rule() == Rule::Tag {
+            let tag_name = element
+                .clone()
```
epage: Do we need this clone?

liquid-compiler/src/parser.rs
```diff
@@ -3,477 +3,831 @@
...
+            iter: next_elements,
+            closed: false,
+        }
+    }
+
+    /// Returns the next element of the block, if any, similarly to an iterator.
+    ///
+    /// However, if the input text reaches its end and the block is not closed,
+    /// an error is returned instead.
+    pub fn next(&mut self) -> Result<Option<BlockElement<'a>>> {
```
epage: What are your thoughts on swapping `Result` and `Option` so we can `impl Iterator`?

Because of `?` it is still handy to have a function that is `Result<Option<_>>`.  That can be a private helper function that the iterator calls into and can just use the [`map_or`](https://doc.rust-lang.org/stable/rust-by-example/error/multiple_error_types/option_result.html) trick to swap them.

src/tags/capture_block.rs
Goncalerta: Yes, `assert_empty` is in every block before returning an `Ok` value.

This happened because of the issues I had with ensuring all elements of the block are used. In the beggining, `TagBlock` was passed as a `&mut` to the signature of blocks, which we were both unsatisfied with. That was needed to call `close` at the end of the block. I then implemented a `Drop` trait, soon realizing that was also a no-go because it would fire even in case of an `Err`: this meant that if something went wrong during parsing and the function was interrupted before finishing parsing (returning an `Err`), the `Drop` trait would see that the iterator wasn't exhausting the elements and `panic!` on top of the `Err`.

Before reverting back to the `&mut` method, I started thinking more about the user side. Besides the signature with `&mut` not being very clear about what the user must expect of the structure, if he forgot by mistake to parse some elements the check would be done silently and perhaps he wouldn't even notice or care about it. By forcing users (with documentation) to use `.assert_empty`, they are in some way reminding themselves that they must exhaust the elements of the block. If they don't wish to parse them (for example in `comment_block`) they can just go through all the elements and skip them (that would happen anyway, it was a matter of being done by the user or off-scene).

I found that this way would be more transparent and clear to understand the use of the API. Of course if a better solution comes along, we can always change this at a later point in time.

src/tags/capture_block.rs
Goncalerta: I don't know if its on every tag/block, I think I just kept that in the ones that already did so in the old parser.

Though I totally think we should do something like that in every tag/block.

liquid-compiler/src/parser.rs
Goncalerta: We did in the past. Now that I look at it, with `Tag` being a structure that already separates the name from tokens, it might be possible to get rid of it, though.

liquid-compiler/src/parser.rs
Goncalerta: I didn't know about that trick with `map_or` :p 

I did the result in the first place because most blocks will want to use `?` because they don't need to handle the errors. This meant that switching them all the time would be tedious and none actually needed the features of an iterator.

For example here: (if_block.rs, line 330)
```
while let Some(element) = tokens.next()? {
```

If `Option`  was first, this kind of syntax would need more boilerplate and it is used in many blocks.

But now that you say it, having two functions, `next()` for iterator and another with the `Result` and `Option` swapped seems the best way to go. What would be a good name for the `Result<Option<_>>` one?

liquid-compiler/src/parser.rs
```diff
@@ -3,477 +3,831 @@
...
-                children.push(t.clone())
-            }
-            options.blocks[x.as_str()].parse(x, &tokens[1..], &children, options)
+    /// Converts the text into a str.
+    pub fn to_str(self) -> &'a str {
+        self.as_str()
+    }
+
+    /// Returns the text as a str.
+    pub fn as_str(&self) -> &'a str {
```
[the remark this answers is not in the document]
Goncalerta: I remember it making sense at the time I wrote it, but right now I honestly can't see why this would be needed. Maybe the code was changed in a way that it stopped being needed or I was thinking wrong.

liquid-compiler/src/parser.rs
```diff
@@ -3,477 +3,831 @@
...
-        let result = result.render(&mut context).unwrap();
-        assert_eq!("42", result);
+        Ok(parse_filter_chain(filterchain))
+    }
+
+    /// Tries to obtain a value from this token.
+    ///
+    /// Do not confuse this value with `liquid-value`'s `Value`.
+    /// In this context, value refers to either a literal value or a variable.
+    pub fn expect_value(mut self) -> std::result::Result<Expression, Self> {
```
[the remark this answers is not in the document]
Goncalerta: So, the problem of expecting something is that sometimes even if one `expect_` function fails you might want to try another. For example, `cycle_tag.rs`, line 58:

```

liquid-compiler/src/parser.rs
epage: True, we have a lot of different cases with a lot of different patterns.  Let's not worry about this right now.

liquid-compiler/src/parser.rs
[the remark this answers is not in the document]
epage: Ok, so this is trying to imitate nom's `IResult`.  It then modifies `self` with the extra information that may be desired in an error.

My concern is that it feels a little dirty.  I try to stick to a rule of the `Result::Err` always having `impl Error`.

So looking into this more:

- Does `expect` taking a `self` buy us anything?  I'm assuming there isn't any allocated memory we're able to reuse.  If it instead took `&self`, then we could make multiple calls without chaining through the error
- Part of the challenge here is getting pest's helpful errors without pest being in the public API.  If we skipped their nice errors, then we wouldn't have a problem.  If instead we exposed `Rule`, then we wouldn't have a problem.  Seems like there is the potential for a middle ground route where we create our own error using the information pest provides us.
- We could switch from `Result` to `Either` or a custom enum.
- We could hold off on any significant change for now to see how the parser evolves.

What are your thoughts?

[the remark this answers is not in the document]
epage: How about we switch to `Either` or a custom enum for now to better convey the intent and worry about changing up error handling later?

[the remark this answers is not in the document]
Goncalerta: I made a custom enum `TryMatchToken`, because I'm not a big fan of `Either`. It seems very ambiguous regarding the meaning of the type.

I think this will take care of this issue (for now).

Goncalerta:
I think this if finally ready :)

epage:
Thanks for all your work on this; this was a monumental task.

PR #241, opened 2018-11-28
Windows line endings

agersant:
When checking out the master branch of liquid-rust on Windows, a number of unit tests were failing due to line-ending differences.

This PR makes Windows users check out the Rust and text files in the depot with Unix line endings to avoid such issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
-*.txt		lf
-*.rs		lf
+*.txt		text eol=lf
+*.rs		text eol=lf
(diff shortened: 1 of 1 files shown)


epage:
Thanks!

While I know this change is minor and probably not needed for this, if you agree to our re-license, could you post on https://github.com/cobalt-org/liquid-rust/issues/7?

agersant:
Done! Would this be worth adding to https://github.com/cobalt-org/liquid-rust/blob/master/CONTRIBUTING.md?

PR #308, opened 2018-12-23
Iterate by proper char indices

badboy:
Not every "char" is just one byte wide. Enumerating `chars()` will lead
to off-by-one (or more) errors with wider characters (e.g. ¹), it could
even panic when accessign a subslice (`&s[last..i]`).

### DIFF
--- a/src/filters/html.rs
+++ b/src/filters/html.rs
@@ -25,7 +25,7 @@ fn _escape(input: &Value, args: &[Value], once_p: bool) -> FilterResult {
     let mut result = String::new();
     let mut last = 0;
     let mut skip = 0;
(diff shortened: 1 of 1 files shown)


epage:
Thanks!

I assume you will need a new release of both liquid and cobalt for this? Would it work if it was delayed a couple more days?  I'm trying to wrap up another major change to liquid.

badboy:
Merry Christmas or happy holidays!

Sure, no rush. I’ll be off for a while anyway ;)

epage:
Merry Christmas!

Thanks for letting me know your schedule.  I'll still try to get this out as soon as I can.

Issue #301, opened 2018-12-12
Revamp Filter API


epage:
# Goals

- Base support for named arguments (for support parser, see https://github.com/cobalt-org/liquid-rust/issues/92)
- Filter writers write signature in declarative way, simplifying error reporting and parameter extraction
- Performance by moving more processing to parse-time
- Allow a tool to generate documentation for all the included filters.
- Offload argument-list parsing error handling from the user

# Steps

## 1. Get filters out of runtime

Modify `FilterCall` to capture the relevant filter, avoiding the need to do this in the runtime

## 2. Change filter APIs

Like tags/blocks, we'll make filters a two-trait process.  The first trait replaces the FilterCall and processes the arguments into the order desired by the filter, erroring if needed.  The second trait is generated by this and is captured into the FilterChain.

We'll also have an introspection trait that reports name, description, and each argument with name, description, and whether it is required, optional, positional only, or keyword only.

Rough sketches
```rust
trait FilterReflection {
    fn name(&self) -> &' static str;
    fn description(&self) -> &'static str;
    // Not sure on return type
    // Goal is to return name and description
    fn required_parameters(&self) -> &'static [(&'static str, &'statci str)];
    fn optional_parameters(&self) -> &'static [(&'static str, &'statci str)];
    // TODO does liquid have positional-only or keyword-only parameters?
    fn positional_parameters(&self) -> (&'static str, &'statci str);
    fn keyword_parameters(&self) -> (&'static str, &'statci str);
}

trait FilterParser: FilterReflection {
    // In the ideal case, this will re-arrange Expressions into a way friendly for `dyn Filter`, reporting errors for invalid combinations of parameters (too few, too many, etc)
    // This will have the added benefit of having moved these checks from Render time to Parse time which is very helpful for application that Render the same template multiple times (like Cobalt)
    // Unsure on exact signature here
    fn parse(&self, Vec<Expression>, HashMap<&str, Expression>) -> Box<dyn Filter>;
}

trait Filter: Debug {
    // This will evaluate the expressions and evaluate the filter.
    fn filter(&self, input: &Value, context: &Contect) -> Value;
}
```

Open Questions
- Need to support positional or keyword only?
- How does liquid-ruby compose a mixture of positional and keyword arguments?

### Making things simple

We'll have a derive macro that takes a Parameters struct and generates an Arguments struct and helpers for taking the arguments and parsing them into the Arguments struct.

[...]

Issue #299, opened 2018-12-11
Allow static analysis of performance


salman-ca:
Just a few general questions on liquid-rust.

1. would you say this implementation provides the same level of safety as the ruby version? i.e. since an end user can create their templates, they won't be able to do anything dangerous.

2. Is there a way to introspect the liquid document.  I believe in the ruby version you can see how many if statements, loops etc. the template has (this is probably to analyze the general complexity of the template - or limit the # of certain tags).

Great work!

epage:
1. I'd say this implementation is safer. The ruby version blurs the line between liquid and ruby and they have to go through extra hoops to ensure it is safe. The Rust version does not have any escape hatches like that.

2. As of right now, there is not a way to introspect the document.  Currently, the document gets turned into `Box<dyn SomeTrait>` where the implementations of the trait aren't exposed such that you can't even dynamic cast them. I'd be interested to know what the use case would be for us to consider opening this up which also help us know what would be the best way to open it up.

salman-ca:
Regarding #2, the use case is as follows.  If you allow the end user to create their own templates, they might write very inefficient templates such as abuse of looping statements etc.
If you can introspect the types and # of statements, you could generate some kind of a usage score on the template (in addition to rendering the template - but introspection could prevent you from evening rendering if it is flagged as such).

That is the use case I am thinking about.

epage:
So a static analysis profiler, interesting.  I'll have to give this some thought on the best way to handle it.

On a related note, we also need to take care of https://github.com/cobalt-org/liquid-rust/issues/300.

salman-ca:
Not sure if it is related to #300 but I know the liquid ruby also has an option where it will render the template and ignore any errors it encounters for some of the tags etc.

epage:
> Not sure if it is related to #300 but I know the liquid ruby also has an option where it will render the template and ignore any errors it encounters for some of the tags etc.

Yeah, currently liquid-rust is aiming to act as if `strict`, `strict_variables`, and `strict_filters` are all enabled.

I get the feeling they regret how lax everything is

> It is recommended that you enable :strict or :warn mode on new apps to stop invalid templates from being created. It is also recommended that you use it in the template editors of existing apps to give editors better error messages.

From https://github.com/Shopify/liquid

Personally, I prefer the strictness to help catch issues in my templates (rather than relying on hand inspecting every feature).  Are you bringing this up because you have a use case where you'd want things more lax?  Could you describe it?

[...]

Issue #267, opened 2018-12-01
`uniq` needs to accept an extra argument?


epage:
Example test
```rust
    assert_eq!(v!([]), filters!(uniq, v!([]), v!("a")));
```

Goncalerta:
See #333 #334 #335

PR #319, opened 2018-12-27
Revamp Filter API [WIP]

Goncalerta:
#301
So in the past days I've been a little distant from working on this project because of all the stuff related to the festivities of this time of the year. Still, I've not been completely away from this project.

I've been trying to learn about procedural macros (apparently, its documentation is a little scarce, at least for someone who is searching it for the first time), as I never had any contact with creating one before.

What I have done here is more of a test to get practice with proc macros. Probably I didn't follow the best practices to implement the macros, so I think it's best to open this PR so you can analyze my attempt and point me in the right direction.

It's very incomplete (doesn't even support keyword parameters) and there could be bugs I'm not aware of that might break the macro in certain situations.


epage:
> So in the past days I've been a little distant from working on this project because of all the stuff related to the festivities of this time of the year. Still, I've not been completely away from this project.

I fully endorse that!  Hope you had a good time.  

> Probably I didn't follow the best practices to implement the macros, so I think it's best to open this PR so you can analyze my attempt and point me in the right direction.

Well, I'll see what I can do but I've never messed with them.  I'm barely getting into macros-by-example.  I've tried to hold off on using macros unless absolutely needed.  I just saw this as a good opportunity and thought you'd enjoy the challenge.

liquid-compiler/src/grammar.pest
```diff
@@ -29,9 +29,11 @@ Variable = ${ Identifier
               )*
             }
 Value = { Literal | Variable }
-Filter = { Identifier ~ (":" ~ Value ~ ("," ~ Value)*)? }
+Filter = { Identifier ~ (":" ~ FilterArgument ~ ("," ~ FilterArgument)*)? }
 FilterChain = { Value ~ ("|" ~ Filter)* }
-
+PositionalFilterArgument = {Value}
+KeywordFilterArgument = {Identifier ~ ":" ~ Value}
+FilterArgument = _{PositionalFilterArgument | KeywordFilterArgument}
```
epage: I leave it to you but I'd recommend holding off on actually implementing #92 to a separate PR to keep this one more focused.

liquid-compiler/src/parser.rs
```diff
@@ -175,14 +176,39 @@ fn parse_value(value: Pair) -> Expression {
...
     if filter.as_rule() != Rule::Filter {
         panic!("Expected a filter.");
     }
 
     let mut filter = filter.into_inner();
     let name = filter.next().expect("A filter always has a name.").as_str();
-    let args = filter.map(parse_value).collect();
+
+    let mut keyword_args = HashMap::new();
+    let mut positional_args = Vec::new();
```
epage: Alternatives: 
- `Vec<(Option<&str>, &str)>`
- `(Vec<&str>, Vec<&str, &str>)`

I'm not sure what having positional arguments after keyword arguments should do.  Error? Actually work with ordering being important?  It might be useful for us to preserve the order of keyword or all arguments. 

I'm also assuming we'll just iterate over these rather than doing lookups, which would be better tuned for `Vec`.

Thoughts?

liquid-compiler/src/parser.rs
```diff
@@ -194,12 +220,27 @@ fn parse_filter(filter: Pair, options: &Language) -> Result<FilterCall> {
...
-    let f = FilterCall::new(name, f, args);
+        })?;
+    let f = f.parse(args)?;
+    
     Ok(f)
 }
 
+// TODO change HashMap and Vec into wrapped structs
+pub struct FilterArguments<'a> {
+    pub positional: Box<Iterator<Item=Expression>>,
```
epage: Boxed iterator or just use `Vec`?  I had originally assumed the latter but have no concrete thoughts one way or the other.

liquid-compiler/src/parser.rs
```diff
@@ -194,12 +220,27 @@ fn parse_filter(filter: Pair, options: &Language) -> Result<FilterCall> {
...
+    pub keyword: HashMap<&'a str, Expression>,
+}
+impl<'a> FilterArguments<'a> {
+    pub fn check_args_exhausted(mut self) -> Result<()> {
```
epage: Do we need this?  Should this check just be in the macro or macro helpers?  This doesn't help us with keyword arguments.

liquid-derive/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "liquid-derive"
+version = "0.18.0"
+authors = ["Johann Hofmann <mail@johann-hofmann.com>"]
```
epage: This is all you, feel free to put your name in here

Never have found good guidance on `authors`.  At what point do we add ourselves?  Remove Johann?

src/filters/mod.rs
```diff
@@ -65,45 +67,189 @@ fn check_args_len(
...
+    // Goal is to return name and description
+    fn required_parameters(&self) -> &'static [(&'static str, &'static str)] {
+        &[("offset", "The offset of the slice.")]
+    }
+    fn optional_parameters(&self) -> &'static [(&'static str, &'static str)] {
+        &[("length", "The length of the slice.")]
+    }
+    // TODO does liquid have positional-only or keyword-only parameters?
+    fn positional_parameters(&self) -> &'static [(&'static str, &'static str)] {
+        &[("offset", "The offset of the slice."), ("length", "The length of the slice.")]
```
epage: The idea for `positional_parameters` is "positional-only", like pythons `*args`, so we shouldn't need to duplicate them here.

Goncalerta:
> > So in the past days I've been a little distant from working on this project because of all the stuff related to the festivities of this time of the year. Still, I've not been completely away from this project.
> 
> I fully endorse that! Hope you had a good time.
> 

I did, thank you! I hope you had too.

> > Probably I didn't follow the best practices to implement the macros, so I think it's best to open this PR so you can analyze my attempt and point me in the right direction.
> 
> Well, I'll see what I can do but I've never messed with them. I'm barely getting into macros-by-example. I've tried to hold off on using macros unless absolutely needed. I just saw this as a good opportunity and thought you'd enjoy the challenge.

Well, if that's the case, I'll try to handle it myself. I mean, even if we don't get it completely right on the first go, we can always come back and fix problems in the future.

liquid-compiler/src/parser.rs
Goncalerta: I chose the iterator because I think we actually only ever need to use `next()`. But I think there won't be a big difference regardless of what we choose so if you prefer I can turn it into a `Vec`.

epage: I honestly don't know if there is a reason to go one way or the other.  Just in my head when considering how to implement this, I was thinking of `Vec`. We ended up with different assumptions so I was curious if yours had any reasoning to it unlike mine.

liquid-compiler/src/parser.rs
Goncalerta: It helps as long as they are accessed with `remove`. I've been thinking about removing the `pub` on keyword and positional, and instead have two methods like `next_positional()` and `get_keyword(&str)` so it could be more clear and straightforward what methods to use to access the args.

It could be done in the macro, but in a way I intended to mimic `TagTokenIter::expect_nothing()`. If a user chooses not to use the macro (for whatever reason needs to manually write `FilterParameters::new()`), it would then still be clear that he couldn't forget about that function. I don't think this is gonna be a likely scenario, though, so I can just take it into the macro if you find that is the best option.

liquid-derive/Cargo.toml
Goncalerta: In my opinion you should have added yourself since the moment you took on as an owner of this project (I mean, you do most contributions and you're the one who deploys the new versions of the crate). You can have more than one author listed in the TOML, so there is no need to delete Johann.

liquid-compiler/src/grammar.pest
Goncalerta: I prefer to include them already, because they will be important in determing the design of the code. For example, the `FilterArguments` function, or methods inside `FilterReflection` such as `positional_parameters` and `keyword_parameters`.

I fear that, if we hold off on them, there is a possibility that we overlook some important aspect that would then require rethinking and refactoring the API again (if we get this PR going just to change the API in the next PR, we will basically need to reimplement all filters twice and there are a lot of them).

liquid-compiler/src/parser.rs
Goncalerta: I think I started with your assumption (because of that outdated comment refering to a `Vec`) but for some reason I ended up changing it into an Iterator (this was before the holidays, so I may be forgetting the actual reason). I guess it was because of the use of `next`.

liquid-compiler/src/parser.rs
Goncalerta: In my opinion, the keyword arguments should just be skipped in the positional count (so in `filter: 1, a:2, 3`, 3 would be the 2nd positional arg).

As for the alternatives, wouldn't that make it hard to lookup the keyword arguments?

Another thing that I've been thinking is: is it really needed to have both positional and keyword args in the same filter? Is there any liquid/jekyll/other filter that does that? Maybe we could just allow either positional or keyword arguments inside a filter, exclusively (`FilterArguments` would then be an enum instead).

epage: > In my opinion, the keyword arguments should just be skipped in the positional count (so in filter: 1, a:2, 3, 3 would be the 2nd positional arg).
>
> Another thing that I've been thinking is: is it really needed to have both positional and keyword args in the same filter? Is there any liquid/jekyll/other filter that does that? Maybe we could just allow either positional or keyword arguments inside a filter, exclusively (FilterArguments would then be an enum instead).

See below about semantics

> As for the alternatives, wouldn't that make it hard to lookup the keyword arguments?

That depends on implementation.  You'd need a `match` for the keyword but it makes checking for exhaustion easy.

src/filters/mod.rs
epage: Related:

> In my opinion, the keyword arguments should just be skipped in the positional count (so in filter: 1, a:2, 3, 3 would be the 2nd positional arg).
>
> Another thing that I've been thinking is: is it really needed to have both positional and keyword args in the same filter? Is there any liquid/jekyll/other filter that does that? Maybe we could just allow either positional or keyword arguments inside a filter, exclusively (FilterArguments would then be an enum instead).

The problem is I can't find any documentation on the semantics.  The only things I can find
- https://help.shopify.com/en/themes/liquid/filters/additional-filters#default_pagination
- https://github.com/Shopify/liquid/blob/master/test/integration/filter_test.rb#L140

I'm unsure what Liquid's semantics are.  My guess is they are Ruby's semantics.

I can speak for python's semantics

```python
def function(arg1, arg2, *pos, kwd1, **kwds):
    pass
```
- `arg1` and `arg2` can be referenced by position or name
- `kwd1` can only be referenced by name
- `pos` gobbles up all positional arguments after `arg2`
- `kwds` gobbles up any named arguments besides `arg1`, `arg2`, and `kwd1`.
- Once keyword arguments start, no more positional arguments can be specified

It was with the assumption of Python's semantics that I wrote the proposal (with TODOs about verifying them)

So looking at the Ruby to try to guess whats going on
- Parsing logic: https://github.com/Shopify/liquid/blob/master/lib/liquid/variable.rb#L96
  - Looks like keyword arguments are intermixed with positional arguments
- Preparing arguments for function: https://github.com/Shopify/liquid/blob/master/lib/liquid/variable.rb#L118
  - Looks like kwargs are made the last parameter, if I'm understanding the Ruby
- The actual call
  - https://github.com/Shopify/liquid/blob/master/lib/liquid/variable.rb#L84
  - https://github.com/Shopify/liquid/blob/master/lib/liquid/context.rb#L85
  - https://github.com/Shopify/liquid/blob/master/lib/liquid/strainer.rb#L54
  - So in the end, this depends on Ruby's semantics.

Goncalerta: So I've been doing some tests with Ruby on [Repl.it](https://repl.it/@Pedro_GoncaloGo/Ruby-positional-and-keyword-arguments-test) and I got to some interesting conclusions:
- An argument is either Positional-only or Keyword-only. At least, I wasn't able to find any syntax that would allow a parameter to be given either as a positional or keyword.
  - This means `FilterReflection` could instead have `positional_parameters` and `keyword_parameters`, returning `(name, description, is_optional)`, where `is_optional` could be a bool or maybe an enum.
  - We could take this even further and simply make a `parameters` function that returns `(name, description, positional_or_keyword, is_optional)`.
- Just like python, no positional args are allowed once keyword args start. As you said, though, ruby liquid filters bypass that, by automatically putting all kw-args in the end of the function call.
  - I think we should separate them. If we had something like `Vec<(Option<&str>, &str)>`, it would be harder to iterate on the positional arguments by their order (if they had any named parameter in the middle it would have to be skipped).
- In the case where the function definition only declares positional arguments and not even a single keyword argument, if we pass in keyword arguments, they will all be merged into a hashmap and work as if we passed in a single positional argument (that happens to be a hashmap) instead. This is the closest I could get to make an argument not kw-only or pos-only.
  - It seems to me that it's such a rare case that it might be unintended. I don't think ruby liquid would want us to pass keyword arguments to a filter that asked for an `Object` positional argument, for instance. So I think we should just ignore this edge case.

Given this, you can go to the link I gave above, if you want. It leads directly to a "project" with my test code and some comments about those conclusions I just pointed out. You can play with it and maybe find something I missed)

epage: From this, some thoughts
- It seems like our macros will need field attributes to declare a field as positional or keyword
- Should we allow "catch all" attributes that are `Vec` and `HashMap` in case a filter author wants to do something very custom?  Nevermind, I guess they could skip the macro in this case.

> ` returning (name, description, is_optional), where is_optional could be a bool or maybe an enum.`

Another option is `Option<&'static str>` where the `str` is a description of the default value.  Except will everything have a default?  Hmm.

> We could take this even further and simply make a parameters function that returns `(name, description, positional_or_keyword, is_optional)`.

Not sure what my thoughts are.  If we do this, `positional_or_keyword` should be an `enum` I think.

Goncalerta: > * It seems like our macros will need field attributes to declare a field as positional or keyword

Yes, this is actually where one of my doubts arises. Because the vast majority of filters will have positional parameters (to the point that keyword parameters are not even very well documented in liquid), I was thinking of making positional arguments the default (so, if no attribute was present, it would assume it's a positional argument. If keyword attr was present, it would be a keyword argument).

Do you like that or would you prefer more explicitness, forcing to always use an attribute?

> Not sure what my thoughts are. If we do this, `positional_or_keyword` should be an `enum` I think.

Yeah, that was exactly my intention. The issue I see with this possibility is that we would then have a 4-elements tuple `(&str, &str, ParameterType, bool)`, which could make it hard to understand the meaning of the return type.

> > ` returning (name, description, is_optional), where is_optional could be a bool or maybe an enum.`
> 
> Another option is `Option<&'static str>` where the `str` is a description of the default value. Except will everything have a default? Hmm.

I think this would aggravate the problem with the tuple structure. Having a `(&str, &str, Option<&str>)`  would be pretty confusing.

liquid-compiler/src/parser.rs
Goncalerta: I'm unsure on which of the alternatives would be better for performance:
- Having a `HashMap` where we remove every element and then assert it is empty;
- Having a `HashMap` where we call `.iter()` and use a `match` on the keyword;
- Having a `Vec<(&str, &str)>`, call `.iter()` and use a `match` on the keyword;
- Some other alternative?

Right now I'm starting to be more inclined to `Vec<(&str, &str)>`, but I'm not completely sure.

epage: Removing from the data structure is extra work that we don't need (and requires things to be `mut` more than needed), so I'm leaning towards `match`.

Since we don't need the other properties of a hash map and they have extra creation overhead, I'd lean towards a `Vec` also.

src/filters/mod.rs
epage: > Yes, this is actually where one of my doubts arises. Because the vast majority of filters will have positional parameters (to the point that keyword parameters are not even very well documented in liquid), I was thinking of making positional arguments the default (so, if no attribute was present, it would assume it's a positional argument. If keyword attr was present, it would be a keyword argument).

Having positional be default makes sense.

> Yeah, that was exactly my intention. The issue I see with this possibility is that we would then have a 4-elements tuple (&str, &str, ParameterType, bool), which could make it hard to understand the meaning of the return type.

We could have a struct for this so that the fields are named.  In the python world, I sometimes refer to tuples as anonymous structs to help people understand their difference from lists.  We'd just be making the return type not-anonymous.

liquid-compiler/src/parser.rs
Goncalerta: Another good thing with the iterator is that it consumes each argument after calling `next`, which allows us to see whether or not all args were actually used (for `check_args_exhausted()`)

src/filters/mod.rs
Goncalerta: Keeping positional arguments and keyword arguments in separated functions maybe would be more useful. Because the user of the function will probably need them separated (for example to have two different documentation sections, one for positional and another for keyword). By mixing them we are just "wasting" performance because the user would then have to do the sorting (of positional vs keyword) instead of it being already statically done.

liquid-compiler/src/filter.rs
```diff
@@ -1,101 +1,121 @@
-use std::fmt::{self, Debug};
-
-use liquid_error;
+use super::FilterArguments;
+use liquid_error::Result;
+use liquid_interpreter::Context;
 use liquid_value::Value;
+use std::fmt::Debug;
```
epage: Don't remember if this is the recommended style but I split `use`s up into three blocks

```
use std

use external

use internal
```

liquid-compiler/src/filter.rs
```diff
@@ -1,101 +1,121 @@
...
-}
 
-/// Support cloning of `Box<FilterValue>`.
-pub trait FilterValueClone {
-    /// Cloning of `dyn FilterValue`.
-    fn clone_box(&self) -> Box<FilterValue>;
+pub struct ParameterReflection {
+    name: &'static str,
+    description: &'static str,
+    is_optiona: bool,
```
epage: Typo :)

liquid-compiler/src/filter.rs
```diff
@@ -1,101 +1,121 @@
...
-    /// Filter `input` based on `arguments`.
-    fn filter(&self, input: &Value, arguments: &[Value]) -> FilterResult;
-}
 
-/// Support cloning of `Box<FilterValue>`.
-pub trait FilterValueClone {
-    /// Cloning of `dyn FilterValue`.
-    fn clone_box(&self) -> Box<FilterValue>;
+pub struct ParameterReflection {
+    name: &'static str,
```
epage: Do these fields need to be `pub`?

liquid-compiler/src/filter_chain.rs
```diff
@@ -80,7 +31,7 @@ impl FilterChain {
 
         // apply all specified filters
         for filter in &self.filters {
-            entry = filter.evaluate(context, &entry)?;
+            entry = filter.filter(&entry, context)?;
```
epage: Maybe we should keep the name `evaluate` to align with the other `evaluate`s in liquid.

Thoughts?

liquid-derive/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+}
+
+// Generates the statement to evaluate the field with the given name into a `Value`
+fn generate_evaluate_field(ident: &Ident, is_option: bool) -> proc_macro2::TokenStream {
+    if is_option {
+        quote! {
+            let #ident = match &self.#ident {
+                Some(field) => Some(field.evaluate(context)?.to_owned()),
```
epage: So this turns a filter parameter from a variable to a value?  Why do we `to_owned`?  Can't we just pass the returned reference to the filter?

liquid-compiler/src/filter.rs
Goncalerta: Yes, they do. I completely forgot and as I still hadn't actually used the struct yet I didn't realize they were private.

liquid-derive/src/lib.rs
Goncalerta: Well, in the old code that is what happens, so I thought there wouldn't be much problem with it.
From [`liquid-compiler/src/filter_chain.rs`](https://github.com/cobalt-org/liquid-rust/blob/master/liquid-compiler/src/filter_chain.rs) line 34
```
let arguments: Result<Vec<Value>> = self
            .arguments
            .iter()
            .map(|a| Ok(a.evaluate(context)?.to_owned()))
            .collect();
```

Passing the reference to the filter means turning:
```
struct EvaluatedSliceParameters {
    offset: Value,
    length: Option<Value>,
}
```
into:
```
struct EvaluatedSliceParameters<'a> {
    offset: &'a Value,
    length: Option<&'a Value>,
}
```

Will filters still be able to do everything they need to with a `&Value` instead of a `Value`? Is there any particular reason `Value`s were owned in the old code?

epage: They should be able to still work fine.  I've been slowly going through, removing unnecessary allocations.  I missed this one.

epage:
Looking good!

Goncalerta:
So I tried to make the code in `liquid-derive` a bit more robust and modular. I hope the macro is not so easily breakable now.

One big problem I stumbled is in the path to liquid structures. I decided to use the full path (because it's possible that the user doesn't use `use`), but that leads to a problem.

For example, looking at `liquid::error::Result<Self>`. Inside the liquid crate itself (where mods are located), there is no mod named `liquid` (it is the root of the path), which means that the code wouldn't compile. To circumvent this I did:
```
// This is necessary to ensure paths generated by macro work inside this module
mod liquid {
    pub use ::*;
}
```

Which I'm not sure if it is the ideal solution. What do you think about it?

epage:
btw just stepped through all the derive code.  While I don't know if I could write it from scratch / modify it, I could at least tell the intent of it all.  So looking good :)

> Which I'm not sure if it is the ideal solution. What do you think about it?

Seems fine to me.

I wonder if the rust2018 path stuff fixes this.

epage:
FYI I'm assuming you are just iterating.  Let me know when you are wanting me to look over anything

Goncalerta:
Yeah, I'm still working on the code, I still have to organize `filter_parameters.rs` code and find out whats the best way to create tests for macros.
If you want to have an idea of what the macro looks like right now:
```
#[derive(Debug, FilterParameters)]
struct SliceParameters {
    // mode = "keyword" just for debug purposes
    #[parameter(description = "The offset of the slice.", mode = "keyword")]
    offset: Expression,
    #[parameter(description = "The length of the slice.")]
    length: Option<Expression>,
}

#[derive(Clone, FilterParser)]
#[filter(
    name = "slice",
    description = "Takes a slice of a given string or array.",
    parameters(SliceParameters),
    parsed(SliceFilter)
)]
pub struct SliceFilterParser;

#[derive(Debug)]
pub struct SliceFilter {
    args: SliceParameters,
}
impl Filter for SliceFilter {
    fn evaluate(&self, input: &Value, context: &Context) -> Result<Value> {
        let args = self.args.evaluate(context)?;

        let offset = args
            .offset
            .as_scalar()
            .and_then(Scalar::to_integer)
            .ok_or_else(|| invalid_argument(0, "Whole number expected"))?;
        let offset = offset as isize;

        let length = args
            .length
            .unwrap_or(&Value::scalar(1))
            .as_scalar()
            .and_then(Scalar::to_integer)
            .ok_or_else(|| invalid_argument(0, "Whole number expected"))?;
        if length < 1 {
            return Err(invalid_argument(1, "Positive number expected"));
        }
        let length = length as isize;

        if let Value::Array(ref input) = *input {
            let (offset, length) = canonicalize_slice(offset, length, input.len());
            Ok(Value::array(
                input
                    .iter()
                    .skip(offset as usize)
                    .take(length as usize)
                    .cloned(),
            ))
        } else {
            let input = input.to_str();
            let (offset, length) = canonicalize_slice(offset, length, input.len());
            Ok(Value::scalar(
                input
                    .chars()
                    .skip(offset as usize)
                    .take(length as usize)
                    .collect::<String>(),
            ))
        }
    }
}
```

Goncalerta:
Right now, in the FilterParser, we must use `#[filter(parameters(...))]` and pass in a struct that implements `FilterParameters`. However, some filters don't have parameters, so I'm thinking of making this attribute optional (if no `parameters` is passed in, FilterParser will default to being a filter with no parameters.


On the other hand, Filters has a `args` field for `FilterParameters`:
```
#[derive(Debug)]
pub struct SliceFilter {
    args: SliceParameters,
}
impl Filter for SliceFilter {
    fn evaluate(&self, input: &Value, context: &Context) -> Result<Value> {
        let args = self.args.evaluate(context)?;

        ...
    }
}
```

I've come up with two possibilites for how the Filter struct should behave in the parametersless case:
```
pub struct SliceFilter { 
     args: (),
}
```
`pub struct SliceFilter;`

However this just makes me think how arbitrary it is to have the parameters stored in `args`. Why not, for example:
```
pub struct SliceFilter(SliceParameters);
```
```
pub struct SliceFilter { 
     some_other_name: SliceParameters,
}
```

So I'm not really sure what's the design I should stick to here (macros depend on that because FilterParser will construct a Filter, so it must know how to pass the arguments).

@epage what do you think would be the best solution?

Goncalerta:
On a completely different note, I removed all the (already commented) items referring to ` FnParseFilter` in `compiler/filter.rs`, because it no longer makes sense to convert a single function to `ParseFilter` (a single function does not have reflection).

This means that the type `BoxedFilterParser` has become obsolete. However, I'm still keeping it (as an alias to `Box<FilterParser>`) so as to keep it consistent with tags and blocks (which still use `Boxed____` types).

Do you think I should keep this alias or would you prefer to replace all `BoxedFilterParser`s with `Box<FilterParser>`?

epage:
> So I'm not really sure what's the design I should stick to here (macros depend on that because FilterParser will construct a Filter, so it must know how to pass the arguments).

So my main care about is to not get in the way of a filter author.   One of my areas of care about is being able to write stateful filters, where the `FilterParser` is initialized with data before being passed to `Liquid` and that data needs to get passed to the `Filter`.  This is why my example code in the issue had an `args`.

This means we need to make it so people can write their own logic for creating each struct.  If we provide a convenience people can easily opt-in for the common case of a stateless filter, thats fine; it just should not get in the way of someone writing their own.

From there, if the exact way the stateless stuff works is hidden from the filter developer, then it doesn't matter.

epage:
> This means that the type BoxedFilterParser has become obsolete. However, I'm still keeping it (as an alias to Box<FilterParser>) so as to keep it consistent with tags and blocks (which still use Boxed____ types).

The intention of the type alias / special box was an optimization for the happy path of stateless filters (removing per-filter allocations when cloning a `Vec` of filters).  This optimization took [200ns off of parse/render](https://github.com/cobalt-org/liquid-rust/commit/cbb1d254b0ed15b32674e7688af6c55cee91c125#diff-2c09afcdc3c420ab0678ba9b5e83959c).  The structure has changed since then, so the optimization probably only impacts creating a `Parser` and shouldn't matter anymore (we don't even bother benchmarking creating parsers).

So I'd say get rid of all of it.

Goncalerta:
@epage, I'd like to hear about your opinion on the macros. They are almost finished, so I'd like to know if you have any suggestion, questions, or want to change anything about them, before I started implementing every filter with the new API.

I hope I don't miss anything:


# How does the creation of a filter work

## ParseFilter

`ParseFilter: Send + Sync + ParseFilterClone + FilterReflection`

Filter Parsers must have the following `impl`s:
  - `Clone`, which can be derived by rust's `Clone` derive.
  - `FilterReflection`, which can be derived with `FilterReflection`. This requires the `#[filter(...)]` attribute, however, with the following meta information:
      - `name = "..."` the name of the filter,
      - `description = "..."` the description of the filter,
      - `parameters(...)` the name of a structure that implements `FilterParameters`, which must be in scope. This attribute is only necessary if the filter has any parameters.
  - `ParseFilter`, which can be derived with `ParseFilter`. This requires the `#[filter(...)]` attribute, however, with the following meta information:
      - `parameters(...)` the name of a structure that implements `FilterParameters`, which must be in scope. This attribute is only necessary if the filter has any parameters.
      - `parsed(...)` the name of a structure that implements `Filter`, which must be in scope. This structure must also implement `FromFilterParameters` (if `parameters(...)` is set) or `Default` (in a parameterless filter)

## FilterParameters

`FilterParameters: Sized + FilterParametersReflection + Debug + Display`

FilterParameters is basically a convenience trait for creating filters with derive macro. As so, implementation of `Display`, `FilterParameters` and the evaluated struct are all included in a single `FilterParameters` macro, as it doesn't really make sense to override any of the implementations without overriding the others.

The only other `derive` macro needed for `FilterParameters` is rust's `Debug`.

The `FilterParameters` macro needs some meta information about the struct:
  - Before `struct` declaration, one can use the `#[evaluated(...)]` parameter to override the name of the evaluated struct. This is in order to avoid eventual name collisions.
  - Before each field, it is mandatory to exist a `#[parameter(...)]` attribute, with information that defines that parameter:
      - `rename = "..."`, (OPTIONAL) although the parameter name (name used in liquid) is by default the same as defined in the field (name used in rust), one may need to use this attribute to avoid name collisions with rust keywords.
      - description = "...", (REQUIRED) the description that will be used for FilterReflection. Requiring this enforces good documentation in every filter.
      - mode = "...", (OPTIONAL) whether this is a `"keyword"` or `"positional"` parameter. By default it is positional, as that is the most common case.
      - value = "...", (OPTIONAL) The type of values allowed. When calling `evaluate` on `FilterParameters`, internally casts (and handles error messages) the resulting `Value` into the corresponding type. The following values are allowed:
        - `"any"`, the default. This parameter can be any liquid value
        - `"whole number"`, this parameter must be an integer scalar,
        - `"fractional number"`, this parameter must be a floating point scalar,
        - `"boolean"`, this parameter must be a boolean,
        - `"date"`, this parameter must be a date,
        - `"string"`, this parameter must be a string.
      
FilterParameters structs are not allowed to be Unit (for that you shouldn't even create the struct in the first place) nor Tuple-like (to enforce explicit field names that will become parameter names).

The type of every field must be either `Expression` or `Option<Expression>`. I didn't care to allow aliased names (set with `type` keyword) as those would just make the FilterParameters declaration less explicit and harder to read. Although the derive doesn't directly check whether or not those types are the types that they are supposed to be (`Expression` could be the name of a structure from another crate, for instance), the macro expansion will eventually failt to compile, raise this error and alert the user.

Using `Expression` sets the parameter as `required`. Using `Option<Expression>` sets the parameter as optional. Required positional parameters must not come after optional positional parameters.

## Filter

`Filter: Send + Sync + Debug + Display`

`Filter` must, once again, implement `Debug` (which can be achieved with rust's macro).
Depending on whether or not the filter has parameters, it must also implement `From<FilterParameters>` or `Default`, respectively (due to `ParseFilter`, has seen above). `Default` can also be achieved with rust. The other trait, however, can be derived with `FromFilterParameters`.

The only meta information needed for this derive is to mark the field that holds `FilterParameters` with `#[parameters]`. Only one field can be marked as so and at least one must be marked. Be aware that every other field will need to have a `Default` implementation.

This trait must also implement `Display`. Luckily, there is a derive for that too: `Display_filter`. This notation might seem a little odd. It basically means that this implements the `Display` trait (as opposed to a nonexistent `DisplayFilter` trait) but it is designed for filters. I saw this this notation being referenced in other places, although I don't know if it is considered a rust standard right now.

For this trait, an attribute `#[name = "..."]` with the name of the filter (the same as in FilterParser's name) must be put before the `struct` declaration. Besides, if there is a `FilterParameters` field it must be marked with `#[parameters]` (but there may be none, as opposed to `FromFilterParameters`).

Finally, one must implement `Filter`. This is the filter function itself, so naturaly there is no derive for that.


# Examples

## Slice

```
#[derive(Debug, FilterParameters)]
struct SliceParameters {
    #[parameter(description = "The offset of the slice.", value = "whole number")]
    offset: Expression,

    #[parameter(description = "The length of the slice.", value = "whole number")]
    length: Option<Expression>,
}

#[derive(Clone, ParseFilter, FilterReflection)]
#[filter(
    name = "slice",
    description = "Takes a slice of a given string or array.",
    parameters(SliceParameters),
    parsed(SliceFilter)
)]
pub struct SliceFilterParser;

#[derive(Debug, FromFilterParameters, Display_filter)]
#[name = "slice"]
pub struct SliceFilter {
    #[parameters]
    args: SliceParameters,
}

impl Filter for SliceFilter {
    fn evaluate(&self, input: &Value, context: &Context) -> Result<Value> {
        let args = self.args.evaluate(context)?;

        let offset = args.offset as isize;
        let length = args.length.unwrap_or(1) as isize;

        if length < 1 {
            return Err(liquid_error::Error::with_msg("Invalid argument")
                // .context("position", "1") position is no longer constant because of keyword args
                .context("cause", "Positive number expected."));
        }

        if let Value::Array(input) = input {
            let (offset, length) = canonicalize_slice(offset, length, input.len());
            Ok(Value::array(
                input.iter().skip(offset).take(length).cloned(),
            ))
        } else {
            let input = input.to_str();
            let (offset, length) = canonicalize_slice(offset, length, input.len());
            Ok(Value::scalar(
                input.chars().skip(offset).take(length).collect::<String>(),
            ))
        }
    }
}
```


## Upcase

```
#[derive(Clone, ParseFilter, FilterReflection)]
#[filter(
    name = "upcase",
    description = "Makes the string uppercase.",
    parsed(UpCaseFilter)
)]
pub struct UpCaseFilterParser;

#[derive(Debug, Default, Display_filter)]
#[name = "upcase"]
pub struct UpCaseFilter;

impl Filter for UpCaseFilter {
    fn evaluate(&self, input: &Value, context: &Context) -> Result<Value> {
        let s = input.to_str();
        Ok(Value::scalar(s.to_uppercase()))
    }
}
```

liquid-compiler/src/parser.rs
```diff
@@ -175,31 +176,60 @@ fn parse_value(value: Pair) -> Expression {
...
+            .context("available filters", available)
+    })?;
+    let f = f.parse(args)?;
+
     Ok(f)
 }
 
+/// Structure that holds both the positional and the keyword arguments of
+/// a filter.
+pub struct FilterArguments<'a> {
```
epage: Should this be in with the filter traits?

liquid-derive/src/filter_parameters.rs
```diff
@@ -0,0 +1,780 @@
...
+    Bool,
+    Date,
+    Str,
+}
+
+impl FromStr for FilterParameterType {
+    type Err = String;
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        match s {
+            "any" => Ok(FilterParameterType::Value),
```
epage: imo the macro parameter `value="human description"` should either be `type="rust type"` or `type="liquid value type"`.  Things like "whole number" are to be less jargon-y for the end-user while here our target is a developer and we should use more concrete, discoverable names.

epage:
I knew I could count on you to do this without me needing to check in, and with a thorough, helpful write-up!

Overall, this looks great!  I think we should have test or example code that covers are major use cases
- positional parameters
- named parameters
- stateless filter
- stateful filter
- using reflection

imo that is the next step to verify the viability of the design.

liquid-compiler/src/parser.rs
Goncalerta: The thing is, right now I'm working on the liquid::error messages (i think it is the only actual feature i didn't completely finish, the rest will be tests and code cleanup)

In order to improve the liquid error messages, I will be sending the span of the filter as a private member of this struct

```
pub struct FilterArguments<'a> {
    pub positional: Box<Iterator<Item = Expression>>,
    pub keyword: Box<Iterator<Item = (&'a str, Expression)> + 'a>,
    span: Span<'a>
}
```

With methods that will allow to raise errors with the line/collumn information. However, putting this with the filter traits would mean using pest types outside `parser.rs`. I personally thing it would be best that everything pest-related should stay in that file (and if it gets too big, we could eventually split it into submodules).

liquid-derive/src/filter_parameters.rs
Goncalerta: Yeah I'm not a big fan of those terms either. I didn't really know what I would do here. On one side, using rust types could be confusing because people could erroneously thing that any rust type could be put there. The liquid types I didn't really know what were the 'official' terms for each type. Guess I'll use the names of scalar enum: `any, integer, float, bool, date, str`.

As for `value` vs `type`, I initially also started with `type`. However, because type is a keyword in rust, for some reason it doesn't let me use it inside the macro. I'm not sure if that is a limitation of macros themselves or in the implementation of the `syn` crate.

```
error: expected identifier, found keyword `type`
  --> src\filters\mod.rs:73:59
   |
73 |     #[parameter(description = "The offset of the slice.", type = "whole number")]
   |                                                           ^^^^ expected identifier, found keyword
```

Goncalerta:
@epage could you give me an example of a stateful filter?

liquid-compiler/src/parser.rs
epage: A couple of questions
- Should we be reworking the errors in this PR?
  - My suspicion is to hold off on doing it this PR (1) to keep this PR smaller (its already huge) and (2) so we can start prototyping and iterating on our error solution in less impactful ways, API wise.
- Should we be using `Span` for this?
  - Long term, we want at least a subset of this information available during `render` and not just `parse` which means we'll be needing to move away from `Span` to another way of tracking this state.

liquid-derive/src/filter_parameters.rs
epage: If nothing else, could just be that they wanted the parser to be as context-free as possible.

Yeah, I was thinking scalar enum variants.  Some of the filters take Array and Object a well.  Not sure if you were also planning on value variants or just having thos be "value".

Hmmm, not sure what to do about `type`.  `value` throws me off because I expect a value with it.  

Alternatively, we could switch the derive to be on the `Arguments` and to use the data type the user specifies.  I don't remember why I had it on the `Parameters`.  I think its because I generally hate idents that are generated by macros and so that was the one I wanted more visible?

epage:
>  could you give me an example of a stateful filter?

None exist right now.  The only stateful plugin I can imagine atm is a block (`highlight` for controlling defaults for syntax highlighting).  Maybe some date ones with a default time zone could be wanted?

liquid-compiler/src/parser.rs
Goncalerta: - Ok, with that in mind I will remove this part from the PR
  - My initial intention was to compensate the regression on the error messages that this PR eventually carries (however the current error messages aren't really that good anyway)
- Well, I think we will end up needing some span-like structure in the end regardless of how we implement it.
  - We could easily switch from `Span` in the future, because it was a private member.

liquid-derive/src/filter_parameters.rs
Goncalerta: I decided not to add Array or Object because those are more complex cases. I see this as an extra shortcut to reduce boilerplate in checking the type of the arguments (in a way, almost like to compensate for the extra boilerplate of the new API in general). What I mean is that people can perfectly opt out of using this element (which defaults to "any") and do the check on their own. For complex cases (for example, an argument that can be either a number or a string; an argument that can be an Object) I think there would be too many cases to cover, so the user is responsible to implement them manually. In the future, however, if one of those cases becomes too common, a new shortcut can always be added for it.

Regarding type vs value, I've just had the sudden idea to call it something like arg_type. It would be a little longer but at least it's more clear what it means.

As for switching the derive to be on the `EvaluatedFilterParameters`, I'm not a fan of that idea. The user would then have to refer to the name of a structure that doesn't exist within their code, which would feel confusing in my opinion.

Goncalerta:
> None exist right now. The only stateful plugin I can imagine atm is a block (`highlight` for controlling defaults for syntax highlighting). Maybe some date ones with a default time zone could be wanted?

What I meant to ask is for a concrete example of a hypothetical filter, so I can check if it is feasible with the current macros or if I would have to make adjustments.

liquid-derive/src/filter_parameters.rs
epage: > In the future, however, if one of those cases becomes too common, a new shortcut can always be added for it.

Alternatively, we can take the `structopt` approach and accept a function name to call to process the data.

> As for switching the derive to be on the `EvaluatedFilterParameters`, I'm not a fan of that idea. The user would then have to refer to the name of a structure that doesn't exist within their code, which would feel confusing in my opinion.

Though there is a similar degree of weirdness that the user is accessing fields in a type that they can't see.  This is why I somewhat lean towards rust types so its very explicit and less guess work is involved.

epage:
> What I meant to ask is for a concrete example of a hypothetical filter, so I can check if it is feasible with the current macros or if I would have to make adjustments to this PR.

I'll go with a date formatting filter where the user can set the default time zone (see #332).

liquid-derive/src/filter_parameters.rs
Goncalerta: >Though there is a similar degree of weirdness that the user is accessing fields in a type that they can't see. This is why I somewhat lean towards rust types so its very explicit and less guess work is involved.

On the other hand, this would make it feel like it would accept any rust type, be it i32, i64, i16, usize, etc..., while the intention is in a way to "unwrap" the `Value` struct into de corresponding scalar, returning whatever that scalar has as a type.

Goncalerta:
I've made this example filter to represent stateful filters:
```
#[derive(Clone, Copy)]
enum Mood {
    Happy,
    Neutral,
    Sad,
}

#[derive(Debug, FilterParameters)]
struct TestStatefulFilterParameters {
    #[parameter(description = "", arg_type = "str")]
    arg: Expression,
}

#[derive(Clone, FilterReflection)]
#[filter(
    name = "state",
    description = "Test stateful filters.",
    parameters(TestStatefulFilterParameters)
)]
pub struct TestStatefulFilterParser {
    state: Mood,
}

impl TestStatefulFilterParser {
    pub fn new() -> Self {
        Self {
            state: Mood::Neutral,
        }
    }

    pub fn make_sad(&mut self) {
        self.mood = Mood::Sad;
    }

    pub fn make_happy(&mut self) {
        self.mood = Mood::Happy;
    }

    pub fn make_neutral(&mut self) {
        self.mood = Mood::Neutral;
    }
}

impl ParseFilter for TestStatefulFilterParser {
    fn parse(&self, arguments: FilterArguments) -> Result<Box<Filter>> {
        let args = TestStatefulFilterParameters::from_args(arguments)?;
        let state = self.state;

        Ok(TestStatefulFilter { args, state })
    }
}

#[derive(Debug, Display_filter)]
#[name = "state"]
pub struct TestStatefulFilter {
    #[parameters]
    args: TestStatefulFilterParameters,
    state: Mood,
}

impl Filter for TestStatefulFilter {
    fn evaluate(&self, _input: &Value, context: &Context) -> Result<Value> {
        let args = self.args.evaluate(context)?;

        let result = match self.state {
            Mood::Happy => format!(":-) {} :-)", args.arg),
            Mood::Sad => format!(":-( {} :-(", args.arg),
            Mood::Neutral => format!(":-| {} :-|", args.arg),
        };

        Ok(Value::scalar(result))
    }
}
```

For the date filter, instead of `Mood` it would be some struct that held the information on the timezone. However, how/where is liquid supposed to mutate this state?

After registering the filter, it is consumed by `ParserBuilder::filter()` and converted into a trait object. I can see no way to access the state after this happens.

If the state would only be tweakable before registering the filter, it would be just like having two stateless filters, wouldn't it?

epage:
> However, how/where is liquid supposed to mutate this state?

I'm not expecting liquid to mutate this state.  I am expecting the user to have to explicitly register this filter so they can initialize it to the state they want.

Though we could have self-mutating filters, tags, and blocks for the sake of performance benchmarkingby having them contain a `Rc<Mutex<_>>` and record where the render-pipeline is at (see #299).

> If the state would only be tweakable before registering the filter, it would be just like having two or more stateless filters, wouldn't it? Maybe I'm not getting the actual motivation for stateful filters.

With the date filter example, we wouldn't want to create a filter per timezone but instead accept the timezone.

Or when we apply this design to tags and blocks, we wouldn't want to have a syntax highlight block per [theme](https://github.com/cobalt-org/cobalt.rs/blob/master/src/syntax_highlight/syntect.rs#L92) (or customized in other ways).  Similarly for [sass configuration](https://github.com/cobalt-org/cobalt.rs/blob/master/src/cobalt_model/sass.rs#L44)

Goncalerta:
> > However, how/where is liquid supposed to mutate this state?
> 
> I'm not expecting liquid to mutate this state. I am expecting the user to have to explicitly register this filter so they can initialize it to the state they want.
> 
> Though we could have self-mutating filters, tags, and blocks for the sake of performance benchmarkingby having them contain a `Rc<Mutex<_>>` and record where the render-pipeline is at (see #299).
> 
> > If the state would only be tweakable before registering the filter, it would be just like having two or more stateless filters, wouldn't it? Maybe I'm not getting the actual motivation for stateful filters.
> 
> With the date filter example, we wouldn't want to create a filter per timezone but instead accept the timezone.
> 
> Or when we apply this design to tags and blocks, we wouldn't want to have a syntax highlight block per [theme](https://github.com/cobalt-org/cobalt.rs/blob/master/src/syntax_highlight/syntect.rs#L92) (or customized in other ways). Similarly for [sass configuration](https://github.com/cobalt-org/cobalt.rs/blob/master/src/cobalt_model/sass.rs#L44)

Oh right, so stateful filters are basically to configure their behavior before their registration, not about mutating state. I think I got it now

Goncalerta:
Note that the compact filter doesn't handle its optional argument yet, although it accepts passing it.

```
 /// Remove nulls from an iterable.  For hashes, you can specify which property you
 /// want to filter out if it maps to Null.
 pub fn compact(input: &Value, args: &[Value]) -> FilterResult {
     check_args_len(args, 0, 1)?;

     let array = input
         .as_array()
         .ok_or_else(|| invalid_input("Array expected"))?;

     // TODO optional property parameter

     let result: Vec<_> = array.iter().filter(|v| !v.is_nil()).cloned().collect();

     Ok(Value::array(result))
 }
```

This makes it accidentally pass this test 
```
#[test]
fn test_compact_empty_array() {
    assert_eq!(v!([]), filters!(Compact, v!([]), v!("a")));
}
```
As it accepts the optional argument and (because the array is empty) doesn't have to do anything with it.

Given this, and because accepting an argument which is not yet implemented may lead users to mistake, I'll mark this test as `#[should_panic]` (just as `test_uniq_empty_array` directly above) and don't allow passing any argument to it.

As to open an issue for this, it's somewhat related to #267, because both are about an optional property parameter (actually both filters plus `sort` and `sort_natural` have a `// TODO optional property parameter`. @epage , should I open a separate issue for each or to have them all in #267?

epage:
> As to open an issue for this, it's somewhat related to #267, because both are about an optional property parameter (actually both filters plus `sort` and `sort_natural` have a `// TODO optional property parameter`). @epage , should I open a separate issue for each or to have them all in #267?

I'd go for separate issue.  If nothing else, it makes it easier for people to scan the compatibility issues to see where we fall short.

Plus, fixing one compatibility issue does not fix the other.

Goncalerta:
@epage 

This PR is almost finished, I think the only things left are to add some documentation for the new filter traits and maybe some minor adjustments.

So you might want to give a look at the whole thing and see if you want to adjust something before this is ready to merge.

liquid-compiler/src/filter.rs
```diff
@@ -1,101 +1,254 @@
...
-
-/// Support cloning of `Box<FilterValue>`.
-pub trait FilterValueClone {
-    /// Cloning of `dyn FilterValue`.
-    fn clone_box(&self) -> Box<FilterValue>;
+/// This is the return type in some `FilterReflection` functions.
+pub struct ParameterReflection {
+    pub name: &'static str,
+    pub description: &'static str,
+    pub is_optional: bool,
```
epage: Adding or changing fields to this struct is a breaking change.  Do you remember what we decided about that?

liquid-compiler/src/filter.rs
```diff
@@ -1,101 +1,254 @@
...
+    fn from_args(args: FilterArguments) -> Result<Self>;
+    fn evaluate(&'a self, context: &'a Context) -> Result<Self::EvaluatedFilterParameters>;
 }
 
-impl FilterValue for FnValueFilter {
-    fn filter(&self, input: &Value, arguments: &[Value]) -> FilterResult {
-        (self.filter)(input, arguments)
-    }
+/// Structure that holds the unparsed arguments of a filter, both positional and keyword.
+pub struct FilterArguments<'a> {
```
epage: Thankfully not seeing us evolve this struct, so probably fine as-is?

src/filters/jekyll/array.rs
```diff
@@ -0,0 +1,316 @@
...
+struct PushFilter {
+    #[parameters]
+    args: PushArgs,
+}
+
+impl Filter for PushFilter {
+    fn evaluate(&self, input: &Value, context: &Context) -> Result<Value> {
+        let args = self.args.evaluate(context)?;
+
+        let element = args.element.clone();
```
epage: `args.element` is a reference which is why we need the `clone`, right?

src/filters/std/string/slice.rs
```diff
@@ -0,0 +1,402 @@
...
+        let args = self.args.evaluate(context)?;
+
+        let offset = args.offset as isize;
+        let length = args.length.unwrap_or(1) as isize;
+
+        if length < 1 {
+            return invalid_argument("length", "Positive number expected").into_err();
+        }
+
+        if let Value::Array(input) = input {
```
epage: The downside of categorizing this as `string` is it also supprts `array`.

It is really hard to figure out how to slice/dice the filters :)

liquid-compiler/src/filter.rs
Goncalerta: What we decided was for the `FilterReflection` trait itself, which was to keep it as a trait instead of a struct for it to be extensible.

For the `ParameterReflection`, we didn't actually discuss about what to do. 

Some possibilities:
  - Add a private `var = ()` field and implement a `new` function to initialize the struct. We could add new methods to change eventual new members without a breaking change.
  - Implement new fields in the original `FilterReflection` trait. This would probably become messy, as it would be harder to connect fields to their parameter. This would actually have be the only way to extend the 4 `parameters` struct in the reflection from the #301 draft.
  - Turn this struct into a trait. This would imply creating a new dull structure for every parameter inside each filter, which maybe could be acceptable because we have macros.

liquid-compiler/src/filter.rs
Goncalerta: I'd say so, especially given that in the #301 draft this struct wouldn't even exist. The two type of arguments would be directly in the function declaration:
```
trait FilterParser: FilterReflection {
    fn parse(&self, Vec<Expression>, HashMap<&str, Expression>) -> Box<dyn Filter>;
}
```
The value I see in this struct is actually more related to organize the function declaration and link positional and keyword arguments, in the sense that they are both just arguments and are mostly (always?) passed in to the functions together. So I'm not really seeing it being extended.

src/filters/jekyll/array.rs
Goncalerta: Yes. Removing that `clone` gives:
```
error[E0308]: mismatched types
  --> src\filters\jekyll\array.rs:41:20
   |
41 |         array.push(element);
   |                    ^^^^^^^ expected enum `liquid_value::Value`, found reference
   |
   = note: expected type `liquid_value::Value`
              found type `&liquid_value::Value`
```

src/filters/std/string/slice.rs
Goncalerta: My bad, didn't notice that!

Goncalerta: This made me notice, by the way, that there aren't any unit tests for slice, even in the original filters implmentation.

liquid-compiler/src/filter.rs
epage: You're right, we can always rev this via filer reflection

Goncalerta:
@epage if CI passes I guess this is good to go

epage:
Again, thank you for all of your hard work!

Issue #320, opened 2018-12-28
Error with double curly brackets "}}" inside code blocks in cobalt 0.15.0


uwearzt:
Starting with 0.15.0 i get an error message

```
Error: build command failed
Caused by: Failed to render content for "posts/2011/03/04/sending-apple-push-notifications-with-erlang.html"
Caused by: liquid:    --> 100:27
    |
100 | "{"aps":{"alert":"Message"}}"␊
    |                           ^---
    |
    = expected EOI, Tag, Expression, or Raw
```

 from the code blob:

```
```erlang
2> pushnotification:send("Message").
"{"aps":{"alert":"Message"}}"
```

this works perfectly fine:

```
```erlang
2> pushnotification:send("Message").
"{"aps":{"alert":"Message"} }"
```

i guess the content inside a code block should be completely ignored by cobalt / liquid.

Geobert:
I've encountered this, and `{% raw %}` nor `{% highlight %}` help :-/

Geobert:
It seems that `{% raw %}` is not working at all

epage:
Liquid has been published.  I've got a PR out to update cobalt and will then release.
https://github.com/cobalt-org/cobalt.rs/pull/590

Issue #248, opened 2018-11-30
Line number of errors can be offset by the client.


epage:
Building on #232, when embedded inside of Cobalt, the liquid will come after the frontmatter, for example
```yaml
---
title: Foo
---

# {{ page.title }}

Hello World!
```

As we add line number support, we're going to need to provide a way to handle offsets like this. 

Options

1. Pass the offset to the parser.

2. Modify the errors as they come out of liquid.

Passing the offset seems like it'd be the simplest.

Issue #244, opened 2018-11-30
Strip preceding / trailing spaces on tags that evaluate to blanks?


epage:
Example test case:
```rust
    assert_template_result!("", wrap(" {% comment %} whatever {% endcomment %} "),);
```

This seems strange.  We need to dig in further to understand under what situations the stripping happens

PR #328, opened 2019-01-19
fix(parser): blocks can accept invalid liquid

### LINKED ISSUE #277: `raw` block doesn't ignore enough
Example test
```rust
    assert_template_result!(
        " Foobar {% invalid ",
        "{% raw %} Foobar {% invalid {% endraw %}",
    );
```

Possibly related to #227

Goncalerta:
Fixes #277 

I don't really like this fix as it is a bit messy, but I guess it gets to do its job, so I think we can consider it even if only as a temporary fix, until we find a better way to handle this issue.

What motivated me into having this fixed was that @Geobert and @epage run into this issue, as reported in #320 and #277 , which means that there might be some urgency in getting this at least working.

Blocks will now be free to parse whatever inside them without liquid restrictions, using `escape_liquid` method. This means invalid liquid will be accepted by `{% raw %}` and `{% comment %}`, as well as any future block that wishes to override liquid rules.

This works by creating a new rule, `LaxLiquidFile`, which instead of directly raising any errors, it produces `InvalidLiquid` tokens, but still tries to parse the rest of the file. Only when parsed, these invalid tokens actually raise an error. This means errors can be avoided by never getting to parse the tokens that would be `InvalidLiquid` (which blocks that override liquid do, by using `escape_liquid` or using `as_str` instead of `parse` in a `BlockElement`).

When calling `parse` in an `InvalidLiquid`, however, it reparses the input (using pest) from the line it is on, this time not allowing `InvalidLiquid` and raising the supposed error. 

This error is then manipulated, so as to add the correct line offset of the input. In the future, having #248 implemented, we can replace this manipulation with the use of an offset in `liquid::error::Error`.

I've also thought of some other solutions, but ended up prefering this one:
- Making comment and raw built-ins
  - This would be another temporary solution, and probably more complicated to implement and it would bring more problems than solutions.
- Using pest to parse Element by Element, instead of the whole file
  - I don't know exactly how pest is implemented. I was afraid that it would not be optimized for this use case and that having to call pest to parse after every single element would be a big hit on performance.
  - Line/col information would be even harder to track than in this case.
- Creating a rule `{%block%} ... {%endblock%}`
  - Again, I don't know how pest works but this seemed like a bad idea. Because tags and blocks cannot be distinguished during pest parsing, the parser would basically need to look for a block closer until the end of the file for **every** tag it encountered.

Note: Some tests were changed to work based on reasons I gave in #277 .

### DIFF
--- a/tests/conformance_ruby/tags/raw_tag_test.rs
+++ b/tests/conformance_ruby/tags/raw_tag_test.rs
@@ -12,7 +12,6 @@ fn test_output_in_raw() {
 }
 
 #[test]
-#[should_panic] // liquid-ignore#277
 fn test_open_tag_in_raw() {
     assert_template_result!(
         " Foobar {% invalid ",
--- a/tests/conformance_ruby/tags/standard_tag_test.rs
+++ b/tests/conformance_ruby/tags/standard_tag_test.rs
@@ -18,7 +18,6 @@ fn test_no_transform() {
 }
 
 #[test]
-#[should_panic] // liquid-rust#277
 fn test_has_a_block_which_does_nothing() {
     assert_template_result!(
         "the comment block should be removed  .. right?",
--- a/liquid-compiler/src/grammar.pest
+++ b/liquid-compiler/src/grammar.pest
@@ -23,7 +23,7 @@ Expression = { ExpressionStart ~ WHITESPACE* ~ ExpressionInner ~ WHITESPACE* ~ E
 // Not allowing Tag/Expression Start/End might become a problem
 // for {% raw %}, {% comment %} and other blocks that don't parse
 // the elements inside with liquid: unclosed delimiters won't be accepted
-Raw = @{ (!(TagStart | ExpressionStart | TagEnd | ExpressionEnd) ~ ANY)+ }
+Raw = @{ (!(TagStart | ExpressionStart) ~ ANY)+ }
 
 
 // Inner parsing
--- a/liquid-compiler/src/grammar.pest
+++ b/liquid-compiler/src/grammar.pest
@@ -1,6 +1,12 @@
 WHITESPACE = _{" " | NEWLINE }
 NON_WHITESPACE_CONTROL_HYPHEN = _{ !"-}}" ~ !"-%}" ~ "-" }
+// Lax liquid file won't raise errors. This allows blocks to override
+// liquid rules and parse their content on their own.
+LaxLiquidFile = ${ SOI ~ (Element | InvalidLiquid)* ~ EOI }
 LiquidFile = ${ SOI ~ Element* ~ EOI }
+
+// A token that could not be parsed as valid liquid
+InvalidLiquid = { !Expression ~ ANY }
 
 // Element-level parsing
 Element = _{ Expression | Tag | Raw }
--- a/liquid-compiler/src/parser.rs
+++ b/liquid-compiler/src/parser.rs
@@ -61,8 +61,8 @@ fn error_from_pair(pair: Pair, msg: String) -> Error {
 
 /// Parses the provided &str into a number of Renderable items.
 pub fn parse(text: &str, options: &Language) -> Result<Vec<Box<Renderable>>> {
-    let mut liquid = LiquidParser::parse(Rule::LiquidFile, text)
-        .map_err(convert_pest_error)?
+    let mut liquid = LiquidParser::parse(Rule::LaxLiquidFile, text)
+        .expect("Parsing with Rule::LaxLiquidFile should not raise errors, but InvalidLiquid tokens instead.")
         .next()
         .expect("Unwrapping LiquidFile to access the elements.")
         .into_inner();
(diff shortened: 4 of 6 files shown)


tests/conformance_ruby/tags/raw_tag_test.rs
```diff
@@ -40,7 +39,7 @@ fn test_open_tag_in_raw() {
     );
     assert_template_result!(
         " test {% raw %} {% endraw %}",
-        "{% raw %} test {% raw %} {% {% endraw %}endraw %}",
+        "{% raw %} test {% raw %} {% endraw %}{% endraw %}",
```
epage: Please revert, these tests should match Ruby to ensure we are compliant.  Besides, I don't see why the new version should pass.  It should end processing at the first `endraw`, causing the second `endraw` to fail.

And boy is that a clever way of printing out the raw tag, breaking it up into
- ` test {% raw %} {%`
- `endraw %}`

tests/conformance_ruby/tags/standard_tag_test.rs
```diff
@@ -40,7 +39,7 @@ fn test_has_a_block_which_does_nothing() {
     assert_template_result!("", "{% comment %}{% blabla %}{% endcomment %}");
     assert_template_result!("", "{%comment%}{% endif %}{%endcomment%}");
     assert_template_result!("", "{% comment %}{% endwhatever %}{% endcomment %}");
-    assert_template_result!("", "{% comment %}{% raw %} {{%%%%}}  }} { {% endcomment %} {% comment {% endraw %} {% endcomment %}");
+    assert_template_result!("", "{% comment %}{% raw %} {{%%%%}}  }} { {% endcomment %}{% comment %}{% endraw %} {% endcomment %}");
```
epage: What is this test change buying us?  Is the problem the fact that this results in leftover whitespace?  If so, this is actually masking https://github.com/cobalt-org/liquid-rust/issues/244 and we should be keeping to Ruby's versions of the tests.

tests/conformance_ruby/tags/raw_tag_test.rs
Goncalerta: Another thing I just noticed:
Are `}}` and `%}` tokens supposed to be accepted (assuming there is no matching `{%`)? As in, is `This input which is just raw text with %} in the middle` accepted in ruby liquid?

Because that can be easily fixed. I just assumed that a `}}` or `%}` would be errors when alone.

Goncalerta:
> delegating to the appropriate plugin how a block should be parsed (including when the block ends).

Just out of curiosity, could you expand more on this? I don't really understand what you meant with "including when the block ends". Do you mean that `{% endblock %}` should be the responsibility of the tagblock and, if it wanted, it could override that notation to something else? Like `{& random tokens endcomment #}`

> This will probably require moving away from pest to another parser like nom though. We'll also want to consider other areas we are having problems with the parser and track these cases on #138.

I agree. Pest looked a lot more promising before I ran into the issues of the extensible nature of liquid. Still, I think there is no need for now to make the change, but we could instead wait until there are enough issues that justify making the change.

Goncalerta:
@epage I've reverted the test changes, but before you merge, I'd like to discuss the two tests, because if I get to understand how they are actually supposed to work, I may be able to fix them with very minor tweaksto this PR.

tests/conformance_ruby/tags/raw_tag_test.rs
epage: From my understanding, all processing should be disabled within a `raw` until `endraw` is encountered, including stray tokens, new, unclosed `raw`, etc.

Goncalerta: So this is where all my confusion came from. My eyes were parsing the input string incorrectly.

I guess `%}` and `}}` should also be allowed, having in mind both this test and #320

tests/conformance_ruby/tags/standard_tag_test.rs
epage: I noticed the whitespace change in the diff but somehow missed the `%}`.

Yuck.  It appears comments are still parsed, just the result is discarded
- Tag: https://github.com/Shopify/liquid/blob/master/lib/liquid/tags/comment.rb
- Base class: https://github.com/Shopify/liquid/blob/master/lib/liquid/block.rb

So the `raw` is being processed and everything within the `raw` is being discarded.

The negative to this is you can't have invalid content in your `comment` without putting it in a `raw`.  The positive is at least it means you can comment out `raw`.

Goncalerta: That is odd... It means that after commenting out a piece of code, it will still waste time getting rendered for nothing.

Maybe in the future it would be worthwhile to create an alternative comment block which trully ignores what's inside and have a way for the user to choose between the two alternatives (based on the same system that would allow choosing jekyll or shopify includes). This way people who wished to have liquid behave the same as ruby's counterpart would be able to, and those who'd like to save time by not rendering unused blocks would also have a choice.

Goncalerta: There are also those tests:
```
    assert_template_result!("", "{%comment%}{%blabla%}{%endcomment%}");
    assert_template_result!("", "{% comment %}{% blabla %}{% endcomment %}");
    assert_template_result!("", "{%comment%}{% endif %}{%endcomment%}");
    assert_template_result!("", "{% comment %}{% endwhatever %}{% endcomment %}");
```
How should comment distinguish between undefined tags like `{%blabla%}` and raw blocks?

epage: > That is odd... It means that after commenting out a piece of code, it will still waste time getting rendered for nothing.

They get parsed but the render never gets called.  If you want to comment out a `raw` that contains `comment`, then this is the only way.

> Maybe in the future it would be worthwhile to create an alternative comment block which trully ignores what's inside

This might get confusing when the workaround is just `{% comment %}{% raw %} ... whatever you want {% endraw %}{% comment %}` and is only saving you from having to spend time parsing the stuff in the `raw`.

epage: > There are also those tests:

Looking at the [ruby code](https://github.com/Shopify/liquid/blob/master/lib/liquid/block.rb#L66) my guess is these are subject to the regular unknown tag handling.  By default, the Ruby implementation has lax handling (ignored) but I assume if you turn on strict handling (the only mode liquid-rust supports), you'd get errors.

So my guess is we should expect these to fail.  In the past, I've either modified the test to pass or changed the test to a failing test.  In this case, I think it is intentionally covering an unknown tag.  So if unknown tags in comments should be a parse error, then I feel we should turn these into assertions that parsing failed.

If we aren't confident enough on this, could always open a "Question" issue and document the test with it.

epage:
thanks for the ping!

PR #350, opened 2019-08-29
Feature/modernize

or17191:
- Remove use of deprecated functions
- Use `dyn` for traits as type arguments (inside Box or Arc).

### DIFF
--- a/liquid-compiler/src/block.rs
+++ b/liquid-compiler/src/block.rs
@@ -35,29 +35,29 @@ pub trait ParseBlock: Send + Sync + ParseBlockClone + BlockReflection {
         arguments: TagTokenIter,
         block: TagBlock,
         options: &Language,
-    ) -> Result<Box<Renderable>>;
+    ) -> Result<Box<dyn Renderable>>;
 }
 
 pub trait ParseBlockClone {
-    fn clone_box(&self) -> Box<ParseBlock>;
+    fn clone_box(&self) -> Box<dyn ParseBlock>;
 }
 
 impl<T> ParseBlockClone for T
 where
     T: 'static + ParseBlock + Clone,
 {
-    fn clone_box(&self) -> Box<ParseBlock> {
+    fn clone_box(&self) -> Box<dyn ParseBlock> {
         Box::new(self.clone())
     }
 }
 
-impl Clone for Box<ParseBlock> {
-    fn clone(&self) -> Box<ParseBlock> {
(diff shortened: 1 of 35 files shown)


epage:
Currently our MSRV is 1.34.0 though I'm fully willing to change it to something higher and have no problem adopting 2019 edition.

or17191:
I'm using version 1.34.0 (I'll also add a `rust-toolchain` file to reduce confusion), but even at that version we use deprecated functions and syntax. My goal here is to compile without warnings.

or17191:
I'm using version 1.34.0 (I'll also add a `rust-toolchain` file to reduce confusion), but even at that version we use deprecated functions and syntax. My goal here is to compile without warnings.

epage:
Oh nice, you even saw the `committed` failures and responded to them.  Since these aren't user facing, the `chore` type would be more appropriate than `feat`.  As for the scope, think of it like "subsystem" in Linux which would be a noun.  In this case, pretty much no scope is needed because it applies to the whole code base.  The theme of "modernizing" is something that can be conveyed in the PR as the overarching motivation and I'll do this as a merge commit, preserving that theme.

(this didn't error in the last PR because I found a bug in the tool that I fixed and since it was a single atomic change, I just squashed the PR to ensure it matched the style)

or17191:
Maybe the `committed` conventions could do with more documentation (maybe in `CONTRIBUTING.md`). I had to "reverse-engineer" why the tool failed, so I didn't notice that the scope was optional. I also didn't know what the possible types are, so I used `feat` because I've seen it in earlier commit subjects.

epage:
Sorry about that.  `committed` is relatively new and this is valuable feedback for me to help make this smooth for new contributors.

For reference, that was enforcement of Conventional Style which has a [spec](https://www.conventionalcommits.org/en/v1.0.0-beta.4/).  At minimum I'll add a print statement to that job to point to the spec.

I should add something to `CONTRIBUTING.md`; I just need to figure out the right phrasing.  I try to be cautious in being dogmatic on the commit style.  I know there are a lot of people who git is still relatively new to them and don't feel comfortable rewriting their history.  I don't want to turn them away.  I usually give people the choice of fix it themselves, be mentored, or have me squash their changes.

Unsure how people who are less experienced will react to the `committed` checks failing.  We'll see.

epage:
I posted https://github.com/cobalt-org/liquid-rust/pull/351 and would love to get your feedback on it.

PR #351, opened 2019-08-30
doc(contrib): Guide people to commit style

epage:


or17191:
This looks great (and would have been very helpful).
You can also:
* Add a "TL;DR". Something along the lines of: Start each commit message with `type(scope): '. Type can be `feat|doc|chore|...`. The scope is optional.
* Tell about the `committed` tool, so contributors will know that they can run it manually if they feel like it.

Issue #262, opened 2018-12-01
Sort order of `nil` is incorrect


epage:
Example test
```rust
    assert_eq!(v!([1, 2, 3, 4, nil]), filters!(sort, v!([nil, 4, 3, 2, 1])));
```

- [Where order is defined](https://github.com/cobalt-org/liquid-rust/blob/master/liquid-value/src/values.rs)
- [Relevant tests](https://github.com/cobalt-org/liquid-rust/search?q=liquid+262&unscoped_q=liquid+262)
  - Once this is working, ideally the test will start failing.  We would just need to remove the `#[should_panic]` on it
  - If the test doesn't start failing, please temporarily remove `#[should_panic]`, and find or create an issue for the new failure

Issue #246, opened 2018-11-30
`compact` filter doesn't properly work on hashes


epage:
Example test:
```
    let assigns = v!({
        "words": ["a", nil, "b", nil, "c"],
        "hashes": [{ "a": "A" }, { "a": nil }, { "a": "C" }],
    });

    // Test hashes
    assert_template_result!(
        "A C",
        "{{hashes | compact: 'a' | map: 'a' | join}}",
        assigns
    );
```

- [Rust implementation](https://github.com/cobalt-org/liquid-rust/blob/master/src/filters/mod.rs)
- [Ruby implementation](https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb)
- [Relevant tests](https://github.com/cobalt-org/liquid-rust/search?q=liquid+246&unscoped_q=liquid+246)
  - Once this is working, ideally the test will start failing.  We would just need to remove the `#[should_panic]` on it
  - If the test doesn't start failing, please temporarily remove `#[should_panic]`, and find or create an issue for the new failure

epage:
> This solves issues #335. It also solves some of issue #246, but the filter doesn't yet work on a single hash.

Issue #335, opened 2019-02-17
`compact` needs to accept an optional `property` parameter


Goncalerta:
See #267, #333, #334 
Example test:
```
assert_eq!(v!([]), filters!(Compact, v!([]), v!("a")));
```

Issue #334, opened 2019-02-17
`sort_natural` needs to accept an optional `property` parameter


Goncalerta:
See #267, #333 #335

PR #352, opened 2019-08-31
Feature/standard filter conformance

or17191:

* This PR should fix issues #262, #333, #334, #335, #257, #250.
* It also partially fixes issue #246. 
* The macro introduced here (`as_sequence`) should help fixing #246, #255, #256, #266 (and maybe others). It helps writing filters that take a value and "coerce" it into an array the same way Ruby's [InputIterator](https://github.com/Shopify/liquid/blob/98dfe198e168eef580345842ff49c4f2549e238b/lib/liquid/standardfilters.rb#L457) does.

I know this is a heavy PR, but the issues fixed here are very closely related.

### DIFF
--- a/src/filters/std/array.rs
+++ b/src/filters/std/array.rs
@@ -47,6 +47,31 @@ impl Filter for JoinFilter {
     }
 }
 
+fn nil_safe_compare(a: &Value, b: &Value) -> Option<cmp::Ordering> {
+    match (a, b) {
+        (Value::Nil, Value::Nil) => Some(cmp::Ordering::Equal),
+        (Value::Nil, _) => Some(cmp::Ordering::Greater),
+        (_, Value::Nil) => Some(cmp::Ordering::Less),
+        (a, b) => a.partial_cmp(b),
+    }
+}
+
+fn nil_safe_casecmp_key(value: &Value) -> Option<String> {
+    match value {
+        Value::Nil => None,
+        value => Some(value.to_str().to_lowercase()),
+    }
+}
+
+fn nil_safe_casecmp(a: &Option<String>, b: &Option<String>) -> Option<cmp::Ordering> {
+    match (a, b) {
+        (None, None) => Some(cmp::Ordering::Equal),
+        (None, _) => Some(cmp::Ordering::Greater),
+        (_, None) => Some(cmp::Ordering::Less),
+        (a, b) => a.partial_cmp(b),
+    }
+}
+
 #[derive(Clone, ParseFilter, FilterReflection)]
 #[filter(
     name = "sort",
@@ -67,7 +92,7 @@ impl Filter for SortFilter {
             .as_array()
             .ok_or_else(|| invalid_input("Array expected"))?;
         let mut sorted = array.clone();
-        sorted.sort_by(|a, b| a.partial_cmp(b).unwrap_or(cmp::Ordering::Equal));
+        sorted.sort_by(|a, b| nil_safe_compare(a, b).unwrap_or(cmp::Ordering::Equal));
         Ok(Value::array(sorted))
     }
 }
@@ -93,9 +118,9 @@ impl Filter for SortNaturalFilter {
             .ok_or_else(|| invalid_input("Array expected"))?;
         let mut sorted: Vec<_> = array
             .iter()
-            .map(|v| (v.to_str().to_lowercase(), v.clone()))
+            .map(|v| (nil_safe_casecmp_key(v), v.clone()))
             .collect();
-        sorted.sort_by(|a, b| a.partial_cmp(b).unwrap_or(cmp::Ordering::Equal));
+        sorted.sort_by(|a, b| nil_safe_casecmp(&a.0, &b.0).unwrap_or(cmp::Ordering::Equal));
         let result: Vec<_> = sorted.into_iter().map(|(_, v)| v).collect();
         Ok(Value::array(result))
     }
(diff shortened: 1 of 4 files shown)


src/filters/std/array.rs
```diff
@@ -4,9 +4,20 @@ use liquid_derive::*;
 use liquid_error::Result;
 use liquid_interpreter::Context;
 use liquid_interpreter::Expression;
-use liquid_value::{Object, Scalar, Value};
+use liquid_value::{Scalar, Value};
 use std::cmp;
 
+macro_rules! as_sequence {
```
epage: I wonder if this would be worthwhile to turn into an iterator for `Value`

src/filters/std/array.rs
```diff
@@ -47,57 +58,130 @@ impl Filter for JoinFilter {
     }
 }
 
+fn nil_safe_compare(a: &Value, b: &Value) -> Option<cmp::Ordering> {
```
epage: Should we modify `Value`s comparator instead?

tests/conformance_ruby/filter_test.rs
```diff
@@ -161,11 +159,10 @@ fn test_sort_natural() {
 }
 
 #[test]
-#[should_panic] // liquid-rust#246
 fn test_compact() {
     let assigns = v!({
         "words": ["a", nil, "b", nil, "c"],
-        "hashes": [{ "a": "A" }, { "a": nil }, { "a": "C" }],
+        "hashes": [{ "a": "A" }, { "a": nil }, { "a": "C" }, {}],
```
epage: Was the `{}` added to test another case?

`conformance_ruby` is a direct port of tests from the ruby implementation and I'd prefer to stay aligned with them and instead to add our own tests to [`filters.rs`](https://github.com/cobalt-org/liquid-rust/blob/master/tests/filters.rs) so I can more easily audit for changes in the ruby tests

tests/conformance_ruby/standard_filter_test.rs
```diff
@@ -401,11 +412,17 @@ fn test_uniq_empty_array() {
 }
 
 #[test]
-#[should_panic] // liquid-rust#335
 fn test_compact_empty_array() {
     assert_eq!(v!([]), filters!(Compact, v!([]), v!("a")));
 }
 
+#[test]
+fn test_compact_invalid_property() {
```
epage: Same as above: I'd prefer new tests not to be put in here

epage:
I'm curious, where is it you are using liquid?

src/filters/std/array.rs
or17191: When I thought about how to implement it, I could have gone with one of the following ways:
* Using a boxed dynamic iterator
* Creating an enum of `slice::Iter`, `iter:Once` and `iter::Empty`
* Using a generic function.

I tried to avoid the first two because of the runtime cost (even the enum would have to query itself every invocation of `next`), but I also didn't like the syntactic overhead of the third option. The macro was a nice compromise (at least I think so).

I, since Ruby used `InputIterator` only for filters, I hesitant about embedding this into `Value`.

src/filters/std/array.rs
or17191: This (and `nil_safe_casecmp`) are implemented as separate functions in Ruby's implementation, so I went the same route.

tests/conformance_ruby/standard_filter_test.rs
or17191: I got this test from [here](https://github.com/Shopify/liquid/blob/14cd011cb5eef042af192ad5294c8f0e1ae89f06/test/integration/standard_filter_test.rb#L345), so I think this is a correct place.

tests/conformance_ruby/filter_test.rs
or17191: I thought `compact` should handle missing attributes, but on a second read of the documentation they don't mention this.

We can go with either:
* Undefined behaviour - Don't test this at all
* Error - Check for this in the filter, and add a test in another file.
* Drop such elements - Move the test to `filters.rs`.

I'm good with either way (although I think the last one is best for compatibility). What do you think?

or17191:
I'm using it to learn rust.

or17191:
> I'm curious, where is it you are using liquid?

I'm using it to learn rust.

src/filters/std/array.rs
epage: If its just a couple of places then that might be fine.  I tried to centralize general properties.

I'm not too worry about the `Box` overhead.  The `clone`s far outweigh any difference that the `Box` would have.

src/filters/std/array.rs
epage: Interesting.  I can go either way.

tests/conformance_ruby/filter_test.rs
epage: If they don't mention it and don't test it, then I'm not too worried which route we go with and am fine going with whatever you recommend.

tests/conformance_ruby/standard_filter_test.rs
epage: Looks like we need to do an audit at some point for updating.  I forked the tests about a year ago and they added this test 7 months ago.

epage:
> I'm using it to learn rust.

I'd say you are doing a great job! If there is a specific area you want more experience with, I can find some issue that would be more targeted.

tests/conformance_ruby/filter_test.rs
or17191: I took option 3 and moved the test to `filters.rs`.

epage:
All good for me to merge once I see that AzDO passes?

or17191:
Yes

PR #360, opened 2019-10-19
Bugfix/url filters

### LINKED ISSUE #253: Tests fail because of a date without time is used, do we need to support that?
Example test:
```rust
    assert_eq!(v!("2001-02-03"), filters!(escape, date(2001, 2, 3)));
```

### LINKED ISSUE #268: `url_decode` doesn't seem to be decoding correctly
Example test:
```rust
    assert_eq!(v!("foo bar"), filters!(url_decode, v!("foo+bar")));
```

or17191:

This PR fixes #268.
It also attempts to fix #253, but I'm not sure if this is the correct fix. The behavior of the `url_{encode,decode}` filters relies on Ruby objects having a `.to_s` method.
If we want to fully emulate this for Date objects in rust, we must make `Date` a first class citizen in the `Value` enum. This would require us to rename the current `Date` object to `DateTime`, which I imagine will break all sorts of backwards compatibility.

What do you think we should do?

### DIFF
--- a/src/filters/std/url.rs
+++ b/src/filters/std/url.rs
@@ -46,6 +46,10 @@ struct UrlEncodeFilter;
 
 impl Filter for UrlEncodeFilter {
     fn evaluate(&self, input: &Value, _context: &Context) -> Result<Value> {
+        if input.is_nil() {
+            return Ok(Value::Nil);
+        }
+
         lazy_static! {
             static ref URL_ENCODE_SET: UrlEncodeSet = UrlEncodeSet("".to_owned());
         }
@@ -72,7 +76,11 @@ struct UrlDecodeFilter;
 
 impl Filter for UrlDecodeFilter {
     fn evaluate(&self, input: &Value, _context: &Context) -> Result<Value> {
-        let s = input.to_str();
+        if input.is_nil() {
+            return Ok(Value::Nil);
+        }
+
+        let s = input.to_str().replace("+", " ");
 
         let result = percent_encoding::percent_decode(s.as_bytes())
             .decode_utf8()


tests/conformance_ruby/standard_filter_test.rs
```diff
@@ -120,14 +120,13 @@ fn test_escape_once() {
...
 #[test]
-#[should_panic] // liquid-rust#253
 fn test_url_encode() {
     assert_eq!(
         v!("foo%2B1%40example.com"),
         filters!(UrlEncode, v!("foo+1@example.com"))
     );
     assert_eq!(v!("1"), filters!(UrlEncode, v!(1)));
-    assert_eq!(v!("2001-02-03"), filters!(UrlEncode, date(2001, 2, 3)));
+    assert_eq!(v!("2001-02-03"), filters!(UrlEncode, date_string(2001, 2, 3)));
```
epage: I think I'd like to keep `date` for now.  Compatibility with liquid-rb will require us to add Date support.

(ditto for decode)

epage:
Don't forget to run `cargo fmt --all` with your changes.  Looks like that is why the CI is failing.

tests/conformance_ruby/standard_filter_test.rs
or17191: How do you want to add Date support? Since every place the repository refers to "Date" it means "DateTime", adding Date will require doing one of two "breaking" things:
- Rename the "Date" concept to "DateTime" and add a "Date" concept. The problem with this is that existing code will be legal but will function differently. The current filter architecture restricts this change, since it maps each parameter type to a single value type, so we can't have a "date" parameter type that starts as a "Date", and then falls back to "DateTime".
- Add a "DateOnly" concept to specify "Date" (vs "DateTime"), which is counter-intuitive.

epage: I'd go for renaming `Date` to `DateTime`.  We still have a relatively high-churn API and few users so I'm ok with breaking.

epage:
Thanks for doing this!

or17191:
I'm having some issues with adding `Date` support. I can't find an API that supports date only string parsing and timezones. `DateTime` doesn't support date only parsing, `Date` doesn't support string parsing at all, and `NaiveDate` doesn't support timezones. Which should I use?

epage:
Does a `Date` need a timezone?  Seems like `NaiveDate` would work.

PR #361, opened 2019-10-30
Use `slice::iter` instead of `into_iter` to avoid future breakage

z2oh:
`an_array.into_iter()` currently just works because of the autoref
feature, which then calls `<[T] as IntoIterator>::into_iter`. But
in the future, arrays will implement `IntoIterator`, too. In order
to avoid problems in the future, the call is replaced by `iter()`
which is shorter and more explicit.

A crater run showed that your crate is affected by a potential future 
change. See https://github.com/rust-lang/rust/pull/65819 for more information.

### DIFF
--- a/liquid-value/src/values.rs
+++ b/liquid-value/src/values.rs
@@ -540,7 +540,7 @@ mod test {
             ("alpha".into(), Value::scalar("1")),
             ("beta".into(), Value::scalar(2f64)),
         ]
-        .into_iter()
+        .iter()
         .cloned()
         .collect();
         let a = Value::Object(a);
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
@epage Do you think you could release this fix as 0.19.1? It would be really great if we could get this fix out soon -- particularly, as a minor version bump :)

epage:
Sorry, I thought I had.  Releasing it now.

epage:
liquid-value 0.19.1 is now out. Glad the problem was in there because liquid has breaking changes in it.

LukasKalbertodt:
Great, thanks :)

PR #363, opened 2019-11-29
Add support for a date only type

### LINKED ISSUE #253: Tests fail because of a date without time is used, do we need to support that?

or17191:

Fixes #253

### DIFF
--- a/liquid-derive/src/filter_parameters.rs
+++ b/liquid-derive/src/filter_parameters.rs
@@ -333,6 +333,7 @@ enum FilterParameterType {
     Float,
     Bool,
     DateTime,
+    Date,
     Str,
 }
 
@@ -345,9 +346,9 @@ impl FromStr for FilterParameterType {
             "float" => Ok(FilterParameterType::Float),
             "bool" => Ok(FilterParameterType::Bool),
             "date_time" => Ok(FilterParameterType::DateTime),
-            "date" => Ok(FilterParameterType::DateTime),
+            "date" => Ok(FilterParameterType::Date),
             "str" => Ok(FilterParameterType::Str),
-            _ => Err(format!("Expected one of the following: \"any\", \"integer\", \"float\", \"bool\", \"date\" or \"str\". Found \"{}\".", s)),
+            _ => Err(format!("Expected one of the following: \"any\", \"integer\", \"float\", \"bool\", \"date_time\", \"date\" or \"str\". Found \"{}\".", s)),
         }
     }
 }
@@ -512,6 +513,15 @@ fn generate_evaluate_field(field: &FilterParameter) -> TokenStream {
                     .context("cause", "DateTime expected")
             )?
         },
+        FilterParameterType::Date => quote! {
+            .as_scalar()
+            .and_then(::liquid::value::Scalar::to_date)
+            .ok_or_else(||
+                ::liquid::error::Error::with_msg("Invalid argument")
+                    .context("argument", #liquid_name)
+                    .context("cause", "Date expected")
+            )?
+        },
         FilterParameterType::Str => quote! {
             .to_str()
         },
@@ -660,6 +670,7 @@ fn generate_evaluated_struct(filter_parameters: &FilterParameters) -> TokenStrea
             FilterParameterType::Float => quote! { f64 },
             FilterParameterType::Bool => quote! { bool },
             FilterParameterType::DateTime => quote! { ::liquid::value::DateTime },
+            FilterParameterType::Date => quote! { ::liquid::value::Date },
             FilterParameterType::Str => quote! { ::std::borrow::Cow<'a, str> },
         };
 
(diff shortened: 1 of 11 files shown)


liquid-derive/src/filter_parameters.rs
```diff
@@ -344,9 +345,10 @@ impl FromStr for FilterParameterType {
             "integer" => Ok(FilterParameterType::Integer),
             "float" => Ok(FilterParameterType::Float),
             "bool" => Ok(FilterParameterType::Bool),
+            "date_time" => Ok(FilterParameterType::DateTime),
```
epage: Unsure whether to do `date_time` or `datetime`.  Thoughts?

liquid-value/src/scalar.rs
```diff
@@ -144,6 +164,7 @@ impl<'s> ScalarCow<'s> {
             ScalarCowEnum::Integer(_) => "whole number",
             ScalarCowEnum::Float(_) => "fractional number",
             ScalarCowEnum::Bool(_) => "boolean",
+            ScalarCowEnum::DateTime(_) => "date time",
```
epage: `date time`, `date / time`, or something else?

This is used for things like errors.

liquid-value/src/scalar.rs
```diff
@@ -241,8 +270,8 @@ impl<'s> PartialEq<bool> for ScalarCow<'s> {
     }
 }
 
-impl<'s> PartialEq<Date> for ScalarCow<'s> {
-    fn eq(&self, other: &Date) -> bool {
+impl<'s> PartialEq<DateTime> for ScalarCow<'s> {
```
epage: Not seeing a `Date` version of this.  Should we have one?

liquid-value/src/scalar.rs
```diff
@@ -284,8 +313,8 @@ impl<'s> PartialOrd<bool> for ScalarCow<'s> {
     }
 }
 
-impl<'s> PartialOrd<Date> for ScalarCow<'s> {
-    fn partial_cmp(&self, other: &Date) -> Option<Ordering> {
+impl<'s> PartialOrd<DateTime> for ScalarCow<'s> {
```
epage: Not seeing a `Date` version of this.  Should we have one?

liquid-value/src/scalar.rs
```diff
@@ -338,6 +369,7 @@ fn scalar_eq<'s>(lhs: &ScalarCow<'s>, rhs: &ScalarCow<'s>) -> bool {
         (&ScalarCowEnum::Float(x), &ScalarCowEnum::Integer(y)) => x == (f64::from(y)),
         (&ScalarCowEnum::Float(x), &ScalarCowEnum::Float(y)) => x == y,
         (&ScalarCowEnum::Bool(x), &ScalarCowEnum::Bool(y)) => x == y,
+        (&ScalarCowEnum::DateTime(x), &ScalarCowEnum::DateTime(y)) => x == y,
```
epage: Should we allow equality between `Date` and `DateTime`?

epage:
Thanks for re-enablin tests for #268 !

Also, I updated the PR message to reference the issue this will fix

liquid-value/src/scalar.rs
```diff
@@ -353,13 +385,37 @@ fn scalar_cmp<'s>(lhs: &ScalarCow<'s>, rhs: &ScalarCow<'s>) -> Option<Ordering>
         (&ScalarCowEnum::Float(x), &ScalarCowEnum::Integer(y)) => x.partial_cmp(&(f64::from(y))),
         (&ScalarCowEnum::Float(x), &ScalarCowEnum::Float(y)) => x.partial_cmp(&y),
         (&ScalarCowEnum::Bool(x), &ScalarCowEnum::Bool(y)) => x.partial_cmp(&y),
+        (&ScalarCowEnum::DateTime(x), &ScalarCowEnum::DateTime(y)) => x.partial_cmp(&y),
```
epage: Should we allow `cmp` between `Date` and `DateTime`

(weird I thought I wrote this up already)

liquid-value/src/scalar.rs
or17191: I think `date time` is better, since `date / time` might imply "date or time".

liquid-derive/src/filter_parameters.rs
or17191: I don't have a clear preference here.

or17191:
I have fixed the CI to the best of my ability, but there seem to be some bigger issues with one of the libraries we use.

epage:
> I have fixed the CI to the best of my ability, but there seem to be some bigger issues with one of the libraries we use.

Maintaining an MSRV is a pain when lower level crates don't test against their minimal versions.

Go ahead and add a commit to update the MSRV in `azure-pipelines.yml` to 1.37.  I think that will fix it.

liquid-derive/src/filter_parameters.rs
epage: I'm unsure as well.  I'll go ahead and mark this resolved then.

epage:
Great work on this!

PR #367, opened 2019-12-14
refactor: Relax `'static` lifetime on reflection

Geobert:

### DIFF
--- a/liquid-compiler/src/block.rs
+++ b/liquid-compiler/src/block.rs
@@ -6,17 +6,17 @@ use super::TagBlock;
 use super::TagTokenIter;
 
 pub trait BlockReflection {
-    fn start_tag(&self) -> &'static str;
+    fn start_tag(&self) -> &str;
 
-    fn end_tag(&self) -> &'static str;
+    fn end_tag(&self) -> &str;
 
-    fn description(&self) -> &'static str;
+    fn description(&self) -> &str;
 
-    fn example(&self) -> Option<&'static str> {
+    fn example(&self) -> Option<&str> {
         None
     }
 
-    fn spec(&self) -> Option<&'static str> {
+    fn spec(&self) -> Option<&str> {
         None
     }
 }
--- a/liquid-compiler/src/filter.rs
+++ b/liquid-compiler/src/filter.rs
@@ -37,8 +37,8 @@ pub trait FilterParametersReflection {
 /// it is necessary to use the `#[filter(...)]`  helper attribute. See documentation on
 /// `liquid-derive` for more information.
 pub trait FilterReflection {
-    fn name(&self) -> &'static str;
-    fn description(&self) -> &'static str;
+    fn name(&self) -> &str;
+    fn description(&self) -> &str;
 
     fn positional_parameters(&self) -> &'static [ParameterReflection];
     fn keyword_parameters(&self) -> &'static [ParameterReflection];
--- a/liquid-compiler/src/registry.rs
+++ b/liquid-compiler/src/registry.rs
@@ -5,7 +5,7 @@ type KeysImpl<'a, K, V> = hash_map::Keys<'a, K, V>;
 
 /// Liquid language plugin registry.
 pub struct PluginRegistry<P> {
-    plugins: MapImpl<&'static str, P>,
+    plugins: MapImpl<String, P>,
 }
 
 impl<P> PluginRegistry<P> {
@@ -21,7 +21,7 @@ impl<P> PluginRegistry<P> {
     /// Generally this is used when setting up the program.
     ///
     /// Returns whether this overrode an existing plugin.
-    pub fn register(&mut self, name: &'static str, plugin: P) -> bool {
+    pub fn register(&mut self, name: String, plugin: P) -> bool {
         let old = self.plugins.insert(name, plugin);
         old.is_some()
     }
@@ -73,7 +73,7 @@ macro_rules! delegate_iterator {
             type Item = $item;
             #[inline]
             fn next(&mut self) -> Option<Self::Item> {
-                self.iter.next().map(|s| *s)
+                self.iter.next().map(|s| s.as_str())
             }
             #[inline]
             fn size_hint(&self) -> (usize, Option<usize>) {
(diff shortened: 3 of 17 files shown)


Geobert:
In order to be able to have custom shortcode in antimoine (and thus cobalt a bit later)

liquid-compiler/src/block.rs
```diff
@@ -6,17 +6,17 @@ use super::TagBlock;
 use super::TagTokenIter;
```
epage: Please add a `BREAKING CHANGE` section to your commit message so it gets in the docs.

See https://www.conventionalcommits.org/en/v1.0.0/

liquid-compiler/src/registry.rs
```diff
@@ -73,7 +73,7 @@ macro_rules! delegate_iterator {
             type Item = $item;
             #[inline]
             fn next(&mut self) -> Option<Self::Item> {
-                self.iter.next().map(|s| *s)
```
epage: Allocating on iteration is a bad practice.  We should be able to get away with `&str`

src/parser.rs
```diff
@@ -156,21 +156,23 @@ where
     /// Inserts a new custom block into the parser
     pub fn block<B: Into<Box<dyn compiler::ParseBlock>>>(mut self, block: B) -> Self {
         let block = block.into();
-        self.blocks.register(block.reflection().start_tag(), block);
+        self.blocks
+            .register(&block.reflection().start_tag().to_owned(), block);
```
epage: So we're converting `&str` (start_tag) -> `String` (to_owned) -> `&str` (&) -> `String` (register)?

liquid-compiler/src/filter.rs
```diff
@@ -37,8 +37,8 @@ pub trait FilterParametersReflection {
 /// it is necessary to use the `#[filter(...)]`  helper attribute. See documentation on
 /// `liquid-derive` for more information.
 pub trait FilterReflection {
-    fn name(&self) -> &'static str;
-    fn description(&self) -> &'static str;
+    fn name(&self) -> &str;
```
epage: I'm surprised, didn't you need to update liquid-derive?

liquid-compiler/src/registry.rs
```diff
@@ -21,8 +21,8 @@ impl<P> PluginRegistry<P> {
     /// Generally this is used when setting up the program.
     ///
     /// Returns whether this overrode an existing plugin.
-    pub fn register(&mut self, name: &'static str, plugin: P) -> bool {
-        let old = self.plugins.insert(name, plugin);
+    pub fn register(&mut self, name: &str, plugin: P) -> bool {
+        let old = self.plugins.insert(name.to_string(), plugin);
```
epage: So using `&'static str` was an optimization that saw huge gains when implemented but things have changed since then.

Could you run and post the results of `cargo +nightly bench` in both master and your branch so we can see if this has any negative impact?

If there is, I have an idea for it.  For my changes, I'm thinking of creating a new string type that is smarter than `Cow<'s, str>`.  Something like

```rust
enum RenameMeString<'s> {
    Owned(String),
    Static(&'static str),
    Borrowed(&'s str),
}
```

This will let us preserve our `'static` optimizations will allowing borrowed/owned to be mixed in.

liquid-compiler/src/block.rs
Geobert: True, I'll add that on the final squash

src/parser.rs
Geobert: True, I'll change register to String (and modify a bunch of tests)

liquid-compiler/src/filter.rs
Geobert: Not that I'm aware, everything compile and I can even use liquid in antimoine as it is with this PR (against liquid 0.19.1, not master)

liquid-compiler/src/registry.rs
Geobert: I'll do the bench, didn't know of that command :D awesome!

liquid-compiler/src/filter.rs
epage: https://github.com/cobalt-org/liquid-rust/blob/master/liquid-derive/src/parse_filter/filter_reflection.rs#L31

Weird, Rust must allow you to over-specify the lifetime in trait implementations.

Geobert: Shall I change the one in liquid-derive?

liquid-compiler/src/registry.rs
Geobert: master: https://gist.github.com/Geobert/897517dd4680e9cc64dedf5ca22b1000
relaxed: https://gist.github.com/Geobert/e1f0207f221b1d0be640f4a074b76bde

how do we read the numbers? What the one in parenthesis means?

liquid-compiler/src/filter.rs
epage: If it doesn't require it I'm unsure if its worth it.

liquid-compiler/src/registry.rs
epage: The framework runs each benchmark a bunch of times and returns the average.  The parens represent how much spread there is above and below the average.

One thing I do when comparing numbers is only care when one of the numbers is outside of the range of the other.

Overall  this looks fine.  Looks like I took out all those extra clones that required the original optimization.

liquid-compiler/src/registry.rs
```diff
@@ -21,8 +21,8 @@ impl<P> PluginRegistry<P> {
     /// Generally this is used when setting up the program.
     ///
     /// Returns whether this overrode an existing plugin.
-    pub fn register(&mut self, name: &str, plugin: P) -> bool {
-        let old = self.plugins.insert(name.to_string(), plugin);
+    pub fn register(&mut self, name: String, plugin: P) -> bool {
+        let old = self.plugins.insert(name, plugin);
```
epage: I wish Rust had a Map where it would extract the key from the value.  With that, we wouldn't need to do any of this.

epage:
Looks good!  Go ahead and cleanup the history and post when you're done and I'll merge it.

Geobert:
Forced pushed to fix length of my breaking change message. I don't understand the beta fmt failing though, I've run cargo fmt and nothing changed

epage:
Thanks!

Geobert:
> 
> 
> Thanks!

Thank you for reviewing it! :)

PR #402, opened 2020-04-28
fix(std): Overflow in `truncate`

### LINKED ISSUE #264: `truncate` has overflow problems
`thread 'conformance_ruby::standard_filter_test::test_truncate' panicked at 'attempt to subtract with overflow', src/filters/mod.rs:218:22`

- [Rust implementation](https://github.com/cobalt-org/liquid-rust/blob/master/src/filters/mod.rs)
- [Ruby implementation](https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb)
- [Relevant tests](https://github.com/cobalt-org/liquid-rust/search?q=liquid+264&unscoped_q=liquid+264)

Geobert:
Fixes #264

### DIFF
--- a/crates/lib/src/stdlib/filters/string/truncate.rs
+++ b/crates/lib/src/stdlib/filters/string/truncate.rs
@@ -72,13 +72,15 @@ impl Filter for TruncateFilter {
         let args = self.args.evaluate(runtime)?;
 
         let lenght = args.lenght.unwrap_or(50) as usize;
-
         let truncate_string = args.ellipsis.unwrap_or_else(|| "...".into());
-
-        let l = cmp::max(lenght - truncate_string.len(), 0);
+        let diff = if lenght >= truncate_string.len() {
+            lenght - truncate_string.len()
+        } else {
+            0
+        };
+        let l = cmp::max(diff, 0);
 
         let input_string = input.to_kstr();
-
         let result = if lenght < input_string.len() {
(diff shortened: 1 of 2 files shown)


Geobert:
> error Subject should not be punctuated but found ``` 

I found this one a bit too much, isn't it?

crates/lib/src/stdlib/filters/string/truncate.rs
```diff
@@ -72,13 +72,15 @@ impl Filter for TruncateFilter {
         let args = self.args.evaluate(runtime)?;
 
         let lenght = args.lenght.unwrap_or(50) as usize;
-
         let truncate_string = args.ellipsis.unwrap_or_else(|| "...".into());
-
-        let l = cmp::max(lenght - truncate_string.len(), 0);
+        let diff = if lenght >= truncate_string.len() {
```
epage, line 76: Oh `lenght`.  This is why I want to get `typos` done so I can run it in CI :)

epage:
> > error Subject should not be punctuated but found ```
> 
> I found this one a bit too much, isn't it?

https://github.com/crate-ci/committed/issues/62

crates/lib/src/stdlib/filters/string/truncate.rs
Geobert: Oh, didn't see it was wrong… #french :p I'll fix it :)

epage:
Thanks!

PR #404, opened 2020-04-30
fix(std): Divide by zero should error, not panic

### LINKED ISSUE #251: `divide_by` 0 should error
Example test:
```rust
    assert_render_error!("{{ 1.0 | divided_by: 0.0 | ceil }}");
```

Shopify:
```ruby
    assert_raises(Liquid::FloatDomainError) do
      assert_template_result "4", "{{ 1.0 | divided_by: 0.0 | ceil }}"
    end
```

- [Rust implementation](https://github.com/cobalt-org/liquid-rust/blob/master/src/filters/math.rs)
- [Ruby implementation](https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb)
- [Relevant tests](https://github.com/cobalt-org/liquid-rust/search?q=liquid+251&unscoped_q=liquid+251)

Geobert:
fixes #251

### DIFF
--- a/crates/lib/src/stdlib/filters/math.rs
+++ b/crates/lib/src/stdlib/filters/math.rs
@@ -313,6 +313,16 @@ impl Filter for DividedByFilter {
             .as_scalar()
             .ok_or_else(|| invalid_argument("operand", "Number expected"))?;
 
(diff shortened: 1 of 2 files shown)


epage:
Thanks!

PR #412, opened 2020-06-16
Introduces into_cow_str for KStringCow, KString, and ScalarCow

fasterthanlime:

This is my second attempt at this, it replaces https://github.com/cobalt-org/liquid-rust/pull/411

### DIFF
--- a/crates/core/src/model/scalar/mod.rs
+++ b/crates/core/src/model/scalar/mod.rs
@@ -5,7 +5,7 @@ mod datetime;
 pub(crate) mod ser;
 
 use std::cmp::Ordering;
-use std::fmt;
+use std::{borrow::Cow, fmt};
 
 use kstring::KString;
 use kstring::KStringCow;
@@ -136,6 +136,14 @@ impl<'s> ScalarCow<'s> {
             ScalarCowEnum::Date(ref x) => Some(*x),
             ScalarCowEnum::Str(ref x) => Date::from_str(x.as_str()),
             _ => None,
+        }
+    }
+
+    /// Interpret as a Cow str, borrowing if possible
+    pub fn into_cow_str(self) -> Cow<'s, str> {
+        match self {
+            Self(ScalarCowEnum::Str(x)) => x.into_cow_str(),
+            other => other.into_string().into_cow_str(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


fasterthanlime:
I considered the `fn to_cow_str(&self)` approach instead (taking self by reference instead of consuming it), but since `Cow<'a, str>` either contains a `String` or an `&'str`, it meant an extra clone in the `Owned` case.

By taking ownership we can "simply" extract the owned string - at least I hope that's what's happening in the `BoxedStr` and `FixedStr` cases.

I'm realizing now that no matter what the `KStringCow` variant is, I can *always* borrow for `'s`, since I'm holding a `&'s dyn ValueView` - it's just that the trait objects get in the way. This PR is one way to allow borrowing from ValueView, I'm sure there's other ways.

crates/core/src/model/scalar/mod.rs
```diff
@@ -1036,4 +1044,44 @@ mod test {
...
+        // does not compile
+        /* {
+            fn extract_cow_str<'s>(value: &'s dyn ValueView) -> Cow<'s, str> {
+                // error: cannot return value referencing temporary value:
+                // `to_kstr(&'a self)` returns a `KStringCow<'a>`, rather than
+                // a `KStringCow<'s>`
+                value.as_scalar().to_kstr().into_cow_str()
+            }
+            assert_eq!(is_borrowed(extract_cow_str(&sc)), true);
+        } */
```
epage, line 1077: Let's break this down
```rust
            fn extract_cow_str<'s>(value: &'s dyn ValueView) -> Cow<'s, str> {
                // error: cannot return value referencing temporary value:
                // `to_kstr(&'a self)` returns a `KStringCow<'a>`, rather than
                // a `KStringCow<'s>`
                let o = value.as_scalar();
                let k = o.to_kstr();
                k.into_cow_str()
            }
```
`o`s type is
```
error[E0308]: mismatched types
    --> crates/core/src/model/scalar/mod.rs:1075:21
     |
1075 |                 let () = o;
     |                     ^^ expected enum `std::option::Option`, found `()`
     |
     = note:   expected enum `std::option::Option<model::scalar::ScalarCow<'_>>`
             found unit type `()`
```
So if `to_kstr` borrow's `o` then it won't live long enough

epage: Whats your preference for how this dead code gets handled before merging?

fasterthanlime: I'm not sure. I wish "compile-fail" tests were a standard Rust feature already, but they're not!

Perhaps it's not useful to have this code at all in there - I wanted you to see it on review though. I can delete it and amend+force-push, or you can drop it yourself, whichever is most convenient for you!

epage: If you don't mind deleting and force-pushing, that'd b e great.  Post here when done since github doesn't send notifications for force-pushes.  I'll then merge this and cut a release.

fasterthanlime:
@epage force-pushed and all checks have passed!

fasterthanlime:
Aand closed by mistake, reopening.

epage:
Everything should now be released.

PR #414, opened 2020-07-07
fix(value)!: Use i64 rather than i32

epage:
Pros:
- More general
- Integrates better with SQL
- Doesn't increase `Scalar`size because there are larger types

Cons
- Doesn't fit within an f64.


epage:
FYI @fasterthanlime

PR #417, opened 2020-07-08
feat(model): Support converting from ValueView

epage:
This is useful for dealing with foreign plugins, like Sql, Lua, etc.


epage:
FYI @fasterthanlime

fasterthanlime:
Just skimmed through the Deserializer, looks good to me!

PR #422, opened 2020-08-03
fix(model): DateTime serialization was off

epage:
When I ported this code from cobalt to liquid, I missed that I was using
serialziation methods on the underlying type rather than our own.


epage:
Step to fixing https://github.com/cobalt-org/cobalt.rs/issues/788

PR #423, opened 2020-08-27
feat(lib): Allow for-loop over `nil`

### LINKED ISSUE #294: Suport for-loop over `nil`
Related: #270 

Example test:
```rust
    assert_template_result!(
        "-",
        "{%for item in array%}+{%else%}-{%endfor%}",
        v!({ "array": nil }),
    );
```

- [for-loop code](https://github.com/cobalt-org/liquid-rust/blob/master/src/tags/for_block.rs)
- [Ruby implementation](https://github.com/Shopify/liquid/blob/master/lib/liquid/tags/for.rb)
- [Relevant tests](https://github.com/cobalt-org/liquid-rust/search?q=liquid+294&unscoped_q=liquid+294)

mattcorbin:
fixes #294 

I added a test in addition to removing the expected panic attribute on the conformance test, let me know if I missed anything.

### DIFF
--- a/crates/lib/src/stdlib/blocks/for_block.rs
+++ b/crates/lib/src/stdlib/blocks/for_block.rs
@@ -132,7 +132,7 @@ fn get_array(runtime: &Runtime<'_>, array_id: &Expression) -> Result<Vec<Value>>
             })
             .collect();
         Ok(x)
-    } else if array.is_state() {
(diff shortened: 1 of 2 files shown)


mattcorbin:
@epage I pushed up the fix - I appreciate the feedback, that's much cleaner!

PR #424, opened 2020-10-09
feat(lib): support for variable passing for `include`

### LINKED ISSUE #310: Suport variable passing for include
~~Depends on #275~~

Resources
- [Relevant tests](https://github.com/cobalt-org/liquid-rust/blob/66607e980263d9e704fb34a789e9146124e5629e/tests/conformance_ruby/tags/include_tag_test.rs) (did I tag the 237 when I meant 310?)
- [Rust code](https://github.com/cobalt-org/liquid-rust/blob/master/crates/lib/src/stdlib/tags/include_tag.rs)
- [Ruby code](https://github.com/Shopify/liquid/blob/master/lib/liquid/tags/include.rb)

JosephLing:
This adds it both for Jekyll and Liquid versions of include which have different syntaxes.

Fixes #310

### DIFF
--- a/crates/lib/src/jekyll/include_tag.rs
+++ b/crates/lib/src/jekyll/include_tag.rs
@@ -27,7 +27,7 @@ impl Renderable for Include {
 
                 for (id, val) in &self.vars {
                     helper_vars.insert(
-                        id.to_owned().into(),
+                        id.clone(),
                         val.try_evaluate(scope)
                             .ok_or_else(|| Error::with_msg("failed to evaluate value"))?
                             .into_owned(),
@@ -108,6 +108,8 @@ impl ParseTag for IncludeTag {
             ));
         }
 
+        arguments.expect_nothing()?;
+
         Ok(Box::new(Include { partial, vars }))
     }
 
@@ -148,6 +150,7 @@ mod test {
             match name {
                 "example.txt" => Some(r#"{{'whooo' | size}}{%comment%}What happens{%endcomment%} {%if num < numTwo%}wat{%else%}wot{%endif%} {%if num > numTwo%}wat{%else%}wot{%endif%}"#.into()),
                 "example_var.txt" => Some(r#"{{include.example_var}}"#.into()),
+                "example_multi_var.txt" => Some(r#"{{include.example_var}} {{include.example}}"#.into()),
                 _ => None
             }
         }
@@ -230,6 +233,24 @@ mod test {
             .build();
         let output = template.render(&mut runtime).unwrap();
         assert_eq!(output, "hello");
+    }
+
+    #[test]
+    fn include_mulitple_varaible() {
+        let text = "{% include example_multi_var.txt example_var=\"hello\" example=\"world\" %}";
+        let options = options();
+        let template = parser::parse(text, &options)
+            .map(runtime::Template::new)
+            .unwrap();
+
+        let partials = partials::OnDemandCompiler::<TestSource>::empty()
+            .compile(::std::sync::Arc::new(options))
+            .unwrap();
+        let mut runtime = RuntimeBuilder::new()
+            .set_partials(partials.as_ref())
+            .build();
+        let output = template.render(&mut runtime).unwrap();
+        assert_eq!(output, "hello world");
     }
 
     #[test]
--- a/crates/lib/src/jekyll/include_tag.rs
+++ b/crates/lib/src/jekyll/include_tag.rs
@@ -1,41 +1,37 @@
 use std::io::Write;
 
-use liquid_core::error::ResultLiquidExt;
+use kstring::KString;
 use liquid_core::parser::TryMatchToken;
 use liquid_core::Expression;
 use liquid_core::Language;
 use liquid_core::Renderable;
-use liquid_core::Result;
 use liquid_core::Runtime;
 use liquid_core::ValueView;
+use liquid_core::{error::ResultLiquidExt, Object, Value};
+use liquid_core::{Error, Result};
 use liquid_core::{ParseTag, TagReflection, TagTokenIter};
 
 #[derive(Debug)]
 struct Include {
     partial: Expression,
-    vars: Vec<(String, Expression)>,
+    vars: Vec<(KString, Expression)>,
 }
 
 impl Renderable for Include {
     fn render_to(&self, writer: &mut dyn Write, runtime: &mut Runtime<'_>) -> Result<()> {
         let name = self.partial.evaluate(runtime)?.render().to_string();
 
-        let mut varaibles_evaluated = Vec::new();
-        for (id, expr) in &self.vars {
-            varaibles_evaluated.push((
-                id.to_owned(),
-                expr.try_evaluate(runtime)
-                    .ok_or_else(|| Error::with_msg("failed to evaluate value"))?
-                    .into_owned(),
-            ));
-        }
-
         runtime.run_in_named_scope(name.clone(), |mut scope| -> Result<()> {
-            if !varaibles_evaluated.is_empty() {
+            if !self.vars.is_empty() {
                 let mut helper_vars = Object::new();
 
-                for (id, val) in varaibles_evaluated {
-                    helper_vars.insert(id.into(), val);
+                for (id, val) in &self.vars {
+                    helper_vars.insert(
+                        id.to_owned().into(),
+                        val.try_evaluate(scope)
+                            .ok_or_else(|| Error::with_msg("failed to evaluate value"))?
+                            .into_owned(),
+                    );
                 }
 
                 scope.stack_mut().set("include", Value::Object(helper_vars));
@@ -94,25 +90,25 @@ impl ParseTag for IncludeTag {
 
         let partial = Expression::with_literal(name);
 
-        let mut vars = Vec::new();
+        let mut vars: Vec<(KString, Expression)> = Vec::new();
         while let Ok(next) = arguments.expect_next("") {
-            let id = next.expect_value().into_result()?.to_string();
+            let id = next.expect_identifier().into_result()?.to_string();
 
             arguments
-                .expect_next("expected string")?
+                .expect_next("\"=\" expected.")?
                 .expect_str("=")
-                .into_result_custom_msg("expected '=' to be used for the assignment")?;
+                .into_result_custom_msg("expected \"=\" to be used for the assignment")?;
 
             vars.push((
-                id,
+                id.into(),
                 arguments
-                    .expect_next("expected expression/value")?
+                    .expect_next("expected value")?
                     .expect_value()
                     .into_result()?,
             ));
         }
 
-        Ok(Box::new(Include { partial }))
+        Ok(Box::new(Include { partial, vars }))
     }
 
     fn reflection(&self) -> &dyn TagReflection {
@@ -151,7 +147,7 @@ mod test {
         fn try_get<'a>(&'a self, name: &str) -> Option<borrow::Cow<'a, str>> {
             match name {
                 "example.txt" => Some(r#"{{'whooo' | size}}{%comment%}What happens{%endcomment%} {%if num < numTwo%}wat{%else%}wot{%endif%} {%if num > numTwo%}wat{%else%}wot{%endif%}"#.into()),
-                "example_var.txt" => Some(r#"{{inlcude.example_var}}"#.into()),
+                "example_var.txt" => Some(r#"{{include.example_var}}"#.into()),
                 _ => None
             }
         }
--- a/crates/lib/src/jekyll/include_tag.rs
+++ b/crates/lib/src/jekyll/include_tag.rs
@@ -13,16 +13,39 @@ use liquid_core::{ParseTag, TagReflection, TagTokenIter};
 #[derive(Debug)]
 struct Include {
     partial: Expression,
+    vars: Vec<(String, Expression)>,
 }
 
 impl Renderable for Include {
     fn render_to(&self, writer: &mut dyn Write, runtime: &mut Runtime<'_>) -> Result<()> {
         let name = self.partial.evaluate(runtime)?.render().to_string();
+
+        let mut varaibles_evaluated = Vec::new();
+        for (id, expr) in &self.vars {
+            varaibles_evaluated.push((
+                id.to_owned(),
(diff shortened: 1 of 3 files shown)


epage:
> I went with the syntax based off jekyll docs however the orignal tests where setup to be var: value, var1: value2. Therefore I can implement the jekyll include tag to be like the docs and the standard lib to be with that setup if that is the plan.

Liquid has a native syntax that Jekyll doesn't use and instead provides its own.  We support both.  So what you want to edit for adding this to the Jekyll include is

https://github.com/cobalt-org/liquid-rust/blob/master/crates/lib/src/jekyll/include_tag.rs

(I'm partial to Shopify because it doesn't support bare strings which forces Jekyll to use nested expressions like `{% include {{ page.my_variable }} %}`)

JosephLing:
So just to triple check just add the same feature to the jekyll include tag with the syntax of:
`{ include 'image.html' path="foo.png" size="large" }`

And that should be it?

epage:
To clarify your clarify 👍 

My understanding is that this PR is adding Jekyll's variable passing support to Shopify's include.  My feedback is that we should follow Shopify and Jekyll's syntax rather than create our own hybrid, so this should instead exist in the Jekyll implementation.  I want to respect your time and am fine if you just do this to Jekyll and don't try to add variable passing to the Shopify version of include.

From what I understanding, the syntax for Jekyll would look like
```
{% include image.html path="foo.png" size="large" %}
```

JosephLing:
So add the changes to the include_tag to allow varaible passing for:
- jekyll [jekyll/tags/include_tag.rs](https://github.com/cobalt-org/liquid-rust/blob/master/crates/lib/src/jekyll/include_tag.rs)
- shopify (but to make sure to follow the spec about no bare strings) [add include_tag.rs to the shopify folder](https://github.com/cobalt-org/liquid-rust/tree/master/crates/lib/src/shopify)

And I'll remove the changes done here: [stdlib/tags/include_tag.rs](https://github.com/cobalt-org/liquid-rust/blob/master/crates/lib/src/stdlib/tags/include_tag.rs)

Sorry for the confusion I thought the issue was to implement for the standard library intially 😅

epage:
> So add the changes to the include_tag to allow varaible passing for:
> 
>     * jekyll [jekyll/tags/include_tag.rs](https://github.com/cobalt-org/liquid-rust/blob/master/crates/lib/src/jekyll/include_tag.rs)
> 
>     * shopify (but to make sure to follow the spec about no bare strings) [add include_tag.rs to the shopify folder](https://github.com/cobalt-org/liquid-rust/tree/master/crates/lib/src/shopify)
> 
> 
> And I'll remove the changes done here: [stdlib/tags/include_tag.rs](https://github.com/cobalt-org/liquid-rust/blob/master/crates/lib/src/stdlib/tags/include_tag.rs)
> 
> Sorry for the confusion I thought the issue was to implement for the standard library intially 😅

No, its alright, liquid makes this very confusing :)

The issue is for the stdlib.  I understood your PR as being interested in getting variable passing behavior for the Jekyll syntax though which is why I directed you over to that.  Feel free to do one, the other, or both as long as we keep our syntaxes straight between shopify and jekyll.

crates/lib/src/jekyll/include_tag.rs
```diff
@@ -13,16 +13,39 @@ use liquid_core::{ParseTag, TagReflection, TagTokenIter};
 #[derive(Debug)]
 struct Include {
     partial: Expression,
+    vars: Vec<(String, Expression)>,
```
epage, line 16: Please use `KString`

crates/lib/src/jekyll/include_tag.rs
```diff
@@ -13,16 +13,39 @@ use liquid_core::{ParseTag, TagReflection, TagTokenIter};
...
 }
 
 impl Renderable for Include {
     fn render_to(&self, writer: &mut dyn Write, runtime: &mut Runtime<'_>) -> Result<()> {
         let name = self.partial.evaluate(runtime)?.render().to_string();
+
+        let mut varaibles_evaluated = Vec::new();
+        for (id, expr) in &self.vars {
```
epage, line 24: Any reason you don't insert into the `Object` directly at this point?

crates/lib/src/jekyll/include_tag.rs
```diff
@@ -69,11 +92,26 @@ impl ParseTag for IncludeTag {
...
         };
 
-        // no more arguments should be supplied, trying to supply them is an error
-        arguments.expect_nothing()?;
-
         let partial = Expression::with_literal(name);
 
+        let mut vars = Vec::new();
+        while let Ok(next) = arguments.expect_next("") {
+            let id = next.expect_value().into_result()?.to_string();
```
epage, line 99: Should this be `expect_identifier` rather than `expect_value`?

crates/lib/src/stdlib/tags/include_tag.rs
```diff
@@ -1,17 +1,18 @@
...
 use liquid_core::Runtime;
 use liquid_core::ValueView;
+use liquid_core::{error::ResultLiquidExt, Object, Value};
 use liquid_core::{Error, Result};
 use liquid_core::{ParseTag, TagReflection, TagTokenIter};
 
 #[derive(Debug)]
 struct Include {
     partial: Expression,
+    vars: Vec<(String, Expression)>,
```
epage, line 15: ditto about KString

crates/lib/src/stdlib/tags/include_tag.rs
```diff
@@ -23,11 +24,36 @@ impl Renderable for Include {
                 .into_err();
         }
         let name = value.to_kstr().into_owned();
+
+        let mut varaibles_evaluated = Vec::new();
```
epage, line 28: ditto about Object

crates/lib/src/stdlib/tags/include_tag.rs
```diff
@@ -68,10 +94,25 @@ impl ParseTag for IncludeTag {
 
         let partial = partial.expect_value().into_result()?;
 
-        // no more arguments should be supplied, trying to supply them is an error
-        arguments.expect_nothing()?;
+        let mut vars = Vec::new();
+        while let Ok(next) = arguments.expect_next("") {
+            let id = next.expect_value().into_result()?.to_string();
```
epage, line 99: ditto about `expect_identifier`

crates/lib/src/stdlib/tags/include_tag.rs
```diff
@@ -68,10 +94,25 @@ impl ParseTag for IncludeTag {
...
+
+            arguments
+                .expect_next("expected string")?
+                .expect_str("=")
```
epage, line 103: Based on the test changes below, it seems this isn't the right syntax but should instead the same syntax used for for-loops.
```rust
fn parse_attr(arguments: &mut TagTokenIter<'_>) -> Result<Expression> {
    arguments
        .expect_next("\":\" expected.")?
        .expect_str(":")
        .into_result_custom_msg("\":\" expected.")?;

    arguments
        .expect_next("Value expected.")?
        .expect_value()
        .into_result()
}

        while let Some(token) = arguments.next() {
            match token.as_str() {
                "limit" => limit = Some(parse_attr(&mut arguments)?),
                "offset" => offset = Some(parse_attr(&mut arguments)?),
                "reversed" => reversed = true,
                _ => {
                    return token
                        .raise_custom_error("\"limit\", \"offset\" or \"reversed\" expected.")
                        .into_err();
                }
            }
        }

```

epage (commented):
Again, thanks for doing this and sorry for being slow to respond

crates/lib/src/jekyll/include_tag.rs
JosephLing: Fixed that now, orginally was using `runtime` to evaluate it without realising I could just use `scope` to do so.

crates/lib/src/stdlib/tags/include_tag.rs
JosephLing: Yep I had assumed incorrectly that jekyll and liquid syntax was going to be the same but found a good syntax guide. So will change it to be the same as the for loop.
liquid: https://www.rubydoc.info/github/Shopify/liquid/Liquid/Include#SYNTAX-constant
jekyll: https://jekyllrb.com/docs/includes/

JosephLing:
> 
> 
> Again, thanks for doing this and sorry for being slow to respond

No worries, I have been slow in fixing/working on this myself

crates/lib/src/jekyll/include_tag.rs
```diff
@@ -1,28 +1,47 @@
...
     fn render_to(&self, writer: &mut dyn Write, runtime: &mut Runtime<'_>) -> Result<()> {
         let name = self.partial.evaluate(runtime)?.render().to_string();
+
         runtime.run_in_named_scope(name.clone(), |mut scope| -> Result<()> {
+            if !self.vars.is_empty() {
+                let mut helper_vars = Object::new();
+
+                for (id, val) in &self.vars {
+                    helper_vars.insert(
+                        id.to_owned().into(),
```
epage, line 30: I think you can make this an `id.clone()`

crates/lib/src/stdlib/tags/include_tag.rs
```diff
@@ -23,11 +25,31 @@ impl Renderable for Include {
...
         runtime.run_in_named_scope(name.clone(), |mut scope| -> Result<()> {
+            // if there our additional varaibles creates a include object to access all the varaibles
+            // from e.g. { include 'image.html' path="foo.png" }
+            // then in image.html you could have <img src="{{include.path}}" />
+            if !self.vars.is_empty() {
+                let mut helper_vars = Object::new();
+
+                for (id, val) in &self.vars {
+                    helper_vars.insert(
+                        id.to_owned().into(),
```
epage, line 38: ditto about clone

crates/lib/src/stdlib/tags/include_tag.rs
```diff
@@ -23,11 +25,31 @@ impl Renderable for Include {
...
+                        val.try_evaluate(scope)
+                            .ok_or_else(|| Error::with_msg("failed to evaluate value"))?
+                            .into_owned(),
+                    );
+                }
+
+                scope.stack_mut().set("include", Value::Object(helper_vars));
```
epage, line 45: Based on the test changes and looking at the Ruby implementation, it looks like we need to set the values directly on  the `stack_mut` rather than as a child of `include`
https://github.com/Shopify/liquid/blob/master/lib/liquid/tags/include.rb#L75

tests/conformance_ruby/tags/include_tag_test.rs
```diff
@@ -81,33 +82,41 @@ fn test_include_tag_for() {
 }
 
 #[test]
-#[should_panic] // liquid-rust#237
+fn test_include_tag_with_single_local_variable() {
```
epage, line 85: Unless this was a test copied over from Ruby, this should instead just live with the tag.

tests/conformance_ruby/tags/include_tag_test.rs
```diff
@@ -81,33 +82,41 @@ fn test_include_tag_for() {
...
     );
 }
 
 #[test]
-#[should_panic] // liquid-rust#237
 fn test_include_tag_with_multiple_local_variables() {
     assert_template_result!(
         "Locale: test123 test321",
-        "{% include 'locale_variables' echo1: 'test123', echo2: 'test321' %}",
+        "{% include 'locale_variables' echo1:'test123' echo2:'test321' %}",
```
epage, line 109: It looks like there are still some syntax gaps we need to handle.

Feel free to split this PR into two and we can get the Jekyll half landed immediately.

tests/conformance_ruby/tags/include_tag_test.rs
JosephLing: I renamed `test_include_tag_with_local_variables` to `test_include_tag_with_single_local_variables` to differeniate between testing for a single varaible and multiple. (master source: https://github.com/cobalt-org/liquid-rust/blob/dc171578116c5ba0f0a9546a02b445c7f0e0b568/tests/conformance_ruby/tags/include_tag_test.rs#L84)

I can move the test into the tag if you want though.

epage: Rename or fork?  I'm still seeing a test by that name.

I'd recommend moving it.  The `conformance_ruby` folder is a direct port of tests from the Ruby version of Liquid and is meant to be untouched except for
- Removing panics
- Updating to better reflect upstream

tests/conformance_ruby/tags/include_tag_test.rs
```diff
@@ -18,13 +18,14 @@ impl liquid::partials::PartialSource for TestFileSystem {
...
             "locale_variables" => "Locale: {{echo1}} {{echo2}}".into(),
+            "single_locale_variable" => "Single Locale: {{echo1}}".into(),
             "variant" => "Variant: {{ variant.title }}".into(),
             "nested_template" => {
                 "{% include 'header' %} {% include 'body' %} {% include 'footer' %}".into()
             }
             "body" => "body {% include 'body_detail' %}".into(),
             "nested_product_template" => {
-                "Product: {{ nested_product_template.title }} {%include 'details'%} ".into()
+                "Product= {{ nested_product_template.title }} {%include 'details'%} ".into()
```
epage, line 28: Overzealous regex?

tests/conformance_ruby/tags/include_tag_test.rs
```diff
@@ -81,33 +82,41 @@ fn test_include_tag_for() {
...
+}
+
+#[test]
+#[should_panic]
 fn test_include_tag_with_local_variables() {
     assert_template_result!(
-        "Locale: test123 ",
-        "{% include 'locale_variables' echo1: 'test123' %}",
+        "Locale: test123",
+        "{% include 'locale_variables' echo1:'test123' %}",
```
epage, line 99: There are still whitespace changes.  I'd recommend reverting the file and just remove the `should_panic`s.

tests/conformance_ruby/tags/include_tag_test.rs
JosephLing: ah oops yep

epage:
I am sorry, somehow I lost track of your last update.

Issue #271, opened 2018-12-01
`forloop` should have a `parentloop` variable


epage:
Example test
```rust
    assert_template_result!(
        "1.1 1.2 1.3 2.1 2.2 2.3 ",
        concat!(
            "{% for inner in outer %}{% for k in inner %}",
            "{{ forloop.parentloop.index }}.{{ forloop.index }} ",
            "{% endfor %}{% endfor %}"
        ),
        v!({"outer": [[1, 1, 1], [1, 1, 1]]}),
    );
```

- [for-loop code](https://github.com/cobalt-org/liquid-rust/blob/master/src/tags/for_block.rs)
- [Ruby implementation](https://github.com/Shopify/liquid/blob/master/lib/liquid/tags/for.rb)
- [Relevant tests](https://github.com/cobalt-org/liquid-rust/search?q=liquid+271&unscoped_q=liquid+271)
  - Once this is working, ideally the test will start failing.  We would just need to remove the `#[should_panic]` on it
  - If the test doesn't start failing, please temporarily remove `#[should_panic]`, and find or create an issue for the new failure

PR #425, opened 2020-11-03
Feat: Added 'parentloop' varaible to 'forloop' for the standard lib

JosephLing:
PR for this issue: https://github.com/cobalt-org/liquid-rust/issues/271

I copied two tests from the conformance test suite and one of them I modified as it seems that `Value::Nil` works slightly different. 


epage:
Looks good.  CI is failing because
- We have a conflict between MSRV and Clippy.  #426 should fix this
- Commit message line length which is easy to go over.  I should bump it.

Interesting question on what to do about `nil`.  It seems that you can index anything on `nil`.

PR #438, opened 2021-03-20
feat(lib): Implement Jekyll's version of sort

epage:
Part of #311.  Continuation of #546

### DIFF
--- a/crates/core/src/parser/parser.rs
+++ b/crates/core/src/parser/parser.rs
@@ -124,7 +124,7 @@ fn parse_literal(literal: Pair) -> Value {
 
 /// Parses a `Variable` from a `Pair` with a variable.
 /// This `Pair` must be `Rule::Variable`.
-fn parse_variable(variable: Pair) -> Variable {
+fn parse_variable_pair(variable: Pair) -> Variable {
     if variable.as_rule() != Rule::Variable {
         panic!("Expected variable.");
     }
@@ -163,7 +163,7 @@ fn parse_value(value: Pair) -> Expression {
 
     match value.as_rule() {
         Rule::Literal => Expression::Literal(parse_literal(value)),
-        Rule::Variable => Expression::Variable(parse_variable(value)),
+        Rule::Variable => Expression::Variable(parse_variable_pair(value)),
         _ => unreachable!(),
     }
 }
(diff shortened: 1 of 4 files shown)


epage:
@limeschool : posting this to keep you updated.

limeschool:
Thanks for working on this, @epage! If I was actually capable of writing decent code, I would've made a PR myself.

Apologies if I came off as whiny earlier, and thanks, really.

epage:
Want to finish this?  I'm trying to wrap up another project before I focus on liquid/cobalt (so I can better focus).

The main other thing to know is the function [try_find](https://docs.rs/liquid-core/0.22.0/liquid_core/model/fn.try_find.html)

limeschool:
I'm working on other things right now myself, but I will give it a looking at once I learn how to write tests (I'm really amateur, haha) and once I've looked at how everything works in this crate.

epage:
>  but I will give it a looking at once I learn how to write tests (I'm really amateur, haha) 

The PR includes some of Jekyll's tests being ported with Rust with the remaining ones partially ported and commented out.  Comparing [the original tests](https://github.com/jekyll/jekyll/blob/master/test/test_filters.rb#L1365) with the ones that I already ported should get you most of the way there.

As for the other parts, the surrounding glue is written, just the implementation needs writing.  As a hint`

```
let args = self.args.evaluate(runtime)?;
```
takes `SortArgs` and constructs a derive-constructed struct that has `property: Option<str>, nils: Option<str>`.  So you can just do `args.property` and `args.nils`.

emmyoh:
@epage I have something like this:
```rust
fn safe_property_getter<'a>(value: &'a Value, property: String) -> Value {
    value
        .as_object()
        .and_then(move |obj| {
            if let Some(scalar) = obj.as_scalar() {
                try_find(&property.clone(), &vec![scalar]).map(|v| v.clone().to_value())
            } else {
                obj.get(&property).map(|v| v.to_value())
            }
        })
        .unwrap_or(Value::Nil)
}
```
It can get a top-level property, but not anything deeper (ie, it fails when `property` is of the form `XXX.YYY`).
Have you figured out how to get nested properties of an `ObjectView`?

emmyoh:
Down to two failing tests:
- `raise_exception_when_input_is_nil`
- `return_sorted_by_property_array_with_nils_first`

If you have the time, would you mind taking a look?

PR #451, opened 2021-12-23
fix(liquid-lib): Make LazyStore cache parsed partials

### LINKED ISSUE #450: LazyStore never caches parsed partials
# System Information
liquid-rust version: commit a28fb391615dae5bfa831444e4dbc4ec279ae7a5

# The Issue
If I understand the purpose of `LazyCompiler` and `LazyStore` correctly, they will cache a parsed partial for every partial that is requested. That way, if the same partial is requested later on, it can simply be retrieved from the cache.

However, there seems to be a bug in this caching system, as it appears that parsed templates are never being cached. Specifically, calling `LazyStore::get()` or `LazyStore::try_get()` do not result in the lazy store's `cache` map containing a key for the partial that was just retrieved via the `get()` or `try_get()` method.

## As a Failing Unit Test
Here is a new unit test that fails under the current implementation that should hopefully illustrate the issue:
```rust
    #[test]
    fn test_store_caches_get() {
        let options = Language::empty();
        let store = lazy::LazyStore {
            language: sync::Arc::new(options),
            source: TestSource,
            cache: sync::Mutex::new(Default::default()),
        };

        assert!(
            !store.cache.lock().unwrap().contains_key("example.txt"),
            "The store cache should not contain the key yet."
        );

        // Look up the partial, causing it to be cached
        let _ = store.get("example.txt").unwrap();

        assert!(
            store.cache.lock().unwrap().contains_key("example.txt"),
            "The store cache should now contain the key."
        );
    }
```
In this test case, `TestSource` is a `PartialSource` containing a single mapping of `"example.txt" => Some("Hello Liquid!".into())`.

This test case fails at the assertion stating "The store cache should now contain the key."

# The Fix
The fix for this issue would be to add a line in `LazyStore::get()` and `LazyStore::try_get()` that adds newly-parsed partials to the `LazyStore`'s `cache`.

# The Pull Request
I have made changes that satisfy the above unit test as well as one for `try_get()` and will be submitting them as a pull request shortly.

Kneelawk:
# This Pull Request
This PR makes `LazyStore` add the parsed `Arc<dyn Renderable>` to its cache when `get_or_create()` or `try_get_or_create()` are called. This PR also adds regression tests that make sure a call to `LazyStore::get()` or `LazyStore::try_get()` adds the parsed partial to its cache.

# Related Issues
This PR fixes #450.

### DIFF
--- a/crates/core/src/partials/lazy.rs
+++ b/crates/core/src/partials/lazy.rs
@@ -105,7 +105,7 @@ where
     S: PartialSource,
 {
     fn try_get_or_create(&self, name: &str) -> Option<sync::Arc<dyn Renderable>> {
-        let cache = self.cache.lock().expect("not to be poisoned and reused");
+        let mut cache = self.cache.lock().expect("not to be poisoned and reused");
         if let Some(result) = cache.get(name) {
             result.as_ref().ok().cloned()
         } else {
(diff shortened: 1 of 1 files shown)


epage (approved):
Can't believe I left that out, thanks!

epage:
Fix is released.

Kneelawk:
Just noticed the commit and PR should have been `fix(liquid-core)`.

Kneelawk:
A rebase could probably fix this, but given the fact that the changes have already been released, it might not be worth it.

epage:
Don't worry about it

PR #457, opened 2022-01-27
Update crate version mentioned in README

dgellow:
The README mentions to add `liquid = "0.20"` to Cargo.toml but the current crate version is 0.23.1


epage:
FYI I had updated our release process so it will update it on next release (see [config](https://github.com/cobalt-org/liquid-rust/blob/master/Cargo.toml#L29)).  I was just waiting for that but this can work.

Issue #446, opened 2021-11-08
Should liquid use `time` instead of `chrono` ?


Geobert:
With `chrono` slowly dying and `time` gaining traction, should liquid consider adopting `time` as well?

epage:
Unsure as I haven't been keeping up on whats going on with these two crates

Geobert:
https://www.reddit.com/r/rust/comments/qamgyh/is_the_chrono_crate_unmaintained/

Since, I think somebody as been approved to join de maintainers team, but the issue is still not fixed and I'm loosing confidence in this crate.

`time` has evolved a lot and can do many things `chrono` is doing and seems more maintained.

VuiMuich:
Hi!
Do you have an estimate how big of an effort this move would be?

We are trying to replace `chrono` with `time` ourselves in [leftwm/leftwm](https://github.com/leftwm/leftwm) but we are getting panics, most probable due to conflicts by attempting to compile two versions of `time` into one binary.

I also agree that `chrono` looks not well with a ton of open PRs and the latest commit being last October :disappointed:

epage:
I started playing around with `time`.  Maybe I missed it but I didn't see support for `%` format strings.  We **might** be able to port to `time` without it (can't remember if parsing that is exposed in important areas) but it'll take a lot more work.

Geobert:
`time` use a different format to specify formats: `"[year]-[month]-[day] [hour]:[minute]:[second] [offset_hour sign:mandatory][offset_minute]"` is what you need :)

epage:
I saw it supported that.  The problem is that it doesn't support the other format.  As I said, switching formats increases the porting work.   I also double checked and we do need a parser for the other format (e.g. [`date`](https://shopify.github.io/liquid/filters/date/)).

VuiMuich:
And there is a `parse` crate feature; I haven't had a look at its mechanics yet, though.

epage:
v0.24.0 is now released with this

PR #460, opened 2022-03-02
feat: Replace chrono with time 0.3

Jake-Shadle:

This PR replaces the unmaintained `chrono` crate with `time` crate. One missing feature in `time` compared to `chrono` is that `time` has its own [format description](https://time-rs.github.io/book/api/format-description.html) that is incompatible with `chrono`, and thus liquid templates, so this PR adds a super simple formatter that replicates the `chrono` behavior. In true Rust fashion once it compiled it worked and all of the tests passed, but I will add some additional tests tomorrow for some niche cases that I noticed were missing from the current tests. I also noticed that chrono was lying in a few cases about what it supported (eg. `%Z` to show the time zone name is not actually implemented in chrono) so in those cases I preserved the `chrono` behavior even when it doesn't match up to the format specifiers [supposedly](https://shopify.github.io/liquid/filters/date/) supported by liquid.

Resolves: #446


crates/core/src/model/scalar/datetime.rs
```diff
@@ -1,12 +1,8 @@
-use std::fmt;
-use std::ops;
```
epage, line 2: Side note: I prefer not to collapse `use`s because it increases the risk of conflicts.  Just a side note because its not worth going back and fixing.

(in general, its good to keep existing formatting when going into code)

Jake-Shadle: Yah sorry, that was my OCD kicking in.

crates/core/src/model/scalar/date.rs
```diff
@@ -1,5 +1,6 @@
-use std::fmt;
-use std::ops;
+use std::{convert::TryInto, fmt, ops};
+
+type DateImpl = time::Date;
```
epage, line 3: nit: `DateImpl` was intentionally put below `Date` to make it so people would first see `pub` items

Jake-Shadle: Ahh ok, that makes sense, in my head I was putting it before the first usage in the pub struct.

crates/core/src/model/scalar/date.rs
```diff
@@ -11,35 +12,35 @@ pub struct Date {
...
-
 impl Date {
     /// Makes a new NaiveDate from the calendar date (year, month and day).
     ///
     /// Panics on the out-of-range date, invalid month and/or day.
-    pub fn from_ymd(year: i32, month: u32, day: u32) -> Self {
-        Self::with_chrono(DateImpl::from_ymd(year, month, day))
+    pub fn from_ymd(year: i32, month: u8, day: u8) -> Self {
+        Self {
+            inner: DateImpl::from_calendar_date(year, month.try_into().unwrap(), day).unwrap(),
```
epage, line 21: Should we be providing better panic messages?

crates/core/src/model/scalar/datetime.rs
```diff
@@ -34,54 +29,384 @@ impl DateTime {
...
+    /// See the [chrono::format::strftime](https://docs.rs/chrono/latest/chrono/format/strftime/index.html)
+    /// module on the supported escape sequences.
+    pub fn format(&self, fmt: &str) -> Result<String, fmt::Error> {
+        use std::fmt::Write;
+        let mut s = String::new();
+
+        let mut cursor = 0;
+        let ts = self.inner;
+
+        while let Some(ind) = fmt[cursor..].find('%') {
```
epage, line 64: Can we split this out from our types, working directly on `time` types?  Doesn't have to happen in this PR, but ideally this would be a dedicated crate.

epage:
Thanks for getting this going!

crates/core/src/model/scalar/datetime.rs
Jake-Shadle: Sure. I briefly entertained the idea of proposing an issue to `time` itself for this, but I wanted to put it here first as a kind of POC that can be extracted later, but this code can be moved to its own module/function for easier extraction in the future for sure.

epage: Yes, I fully support starting with a specific case and generalizing out.  I see ensuring this is decoupled as a step to help towards that.  I feel like it'd also make it easier to port test cases over from chrono (I assume they have them) and ensure they continue to work no matter where this lives.

Jake-Shadle:
Ok, I believe this is ready, I've added a bunch of tests, and expanded the implemntation to support all of what Ruby (at least in their documentation) says they support, which is what I assumed should be considered the reference target for what this project supports. Notably, this now supports all of the flags and padding width.

`%<flags><width><modifier><conversion>`

### `<flags>`
- `-`  don't pad a numerical output
- `_`  use spaces for padding
- `0`  use zeros for padding
- `^`  upcase the result string
- `#`  change case
- `:`  use colons for `%z`

### `<width>`
Integer specifying padding width

### `<modifier>`
`E` or `O` modifier, Ruby recognizes them, but ignores them, so this does the same.

### `<conversion>`
#### Date (Year, Month, Day):
- `%Y` - Year with century if provided, will pad result at least 4 digits. -0001, 0000, 1995, 2009, 14292, etc.
- `%C` - year / 100 (rounded down such as 20 in 2009)
- `%y` - year % 100 (00..99)
- `%m` - Month of the year, zero-padded (01..12)
  - `%_m`  blank-padded ( 1..12)
  - `%-m`  no-padded (1..12)
- `%B` - The full month name (``January'')
  - `%^B`  uppercased (``JANUARY'')
- `%b` - The abbreviated month name (``Jan'')
  `%^b`  uppercased (``JAN'')
- `%h` - Equivalent to `%b`
- `%d` - Day of the month, zero-padded (01..31)
  - `%-d`  no-padded (1..31)
- `%e` - Day of the month, blank-padded ( 1..31)
- `%j` - Day of the year (001..366)

#### Time (Hour, Minute, Second, Subsecond):
- `%H` - Hour of the day, 24-hour clock, zero-padded (00..23)
- `%k` - Hour of the day, 24-hour clock, blank-padded ( 0..23)
- `%I` - Hour of the day, 12-hour clock, zero-padded (01..12)
- `%l` - Hour of the day, 12-hour clock, blank-padded ( 1..12)
- `%P` - Meridian indicator, lowercase (``am'' or ``pm'')
- `%p` - Meridian indicator, uppercase (``AM'' or ``PM'')
- `%M` - Minute of the hour (00..59)
- `%S` - Second of the minute (00..60)
- `%L` - Millisecond of the second (000..999) The digits under millisecond are truncated to not produce 1000.
- `%N` - Fractional seconds digits, default is 9 digits (nanosecond) Note that `time` only has a precision of nanoseconds so digits >9 will always be 0.
    - `%3N`  millisecond (3 digits)
    - `%6N`  microsecond (6 digits)
    - `%9N`  nanosecond (9 digits)
    - `%12N` picosecond (12 digits)
    - `%15N` femtosecond (15 digits)
    - `%18N` attosecond (18 digits)
    - `%21N` zeptosecond (21 digits)
    - `%24N` yoctosecond (24 digits)

#### Time zone
- `%z` - Time zone as hour and minute offset from UTC (e.g. +0900)
    - `%:z` - hour and minute offset from UTC with a colon (e.g. +09:00)
    - `%::z` - hour, minute and second offset from UTC (e.g. +09:00:00)
- `%Z` - In Ruby/C, this is Abbreviated time zone name or similar information.  (OS dependent), however this was never supported by `chrono`, and would require added more dependencies since `time` doesn't have this information, so this implementation just does the same conversion that chrono did, which is that `%Z` is equivalent to `%:z`

#### Weekday
- `%A` - The full weekday name (``Sunday'')
  - `%^A`  uppercased (``SUNDAY'')
- `%a` - The abbreviated name (``Sun'')
  - `%^a`  uppercased (``SUN'')
- `%u` - Day of the week (Monday is 1, 1..7)
- `%w` - Day of the week (Sunday is 0, 0..6)

#### ISO 8601 week-based year and week number
The first week of YYYY starts with a Monday and includes YYYY-01-04.
The days in the year before the first week are in the last week of
the previous year.
- `%G` - The week-based year
- `%g` - The last 2 digits of the week-based year (00..99)
- `%V` - Week number of the week-based year (01..53)

#### Week number
The first week of YYYY that starts with a Sunday or Monday (according to %U
or %W). The days in the year before the first week are in week 0.
- `%U` - Week number of the year. The week starts with Sunday. (00..53)
- `%W` - Week number of the year. The week starts with Monday. (00..53)

#### Seconds since the Epoch
- `%s` - Number of seconds since 1970-01-01 00:00:00 UTC.

#### Literal string
- `%n` - Newline character (\n)
- `%t` - Tab character (\t)
- `%%` - Literal ``%'' character

#### Combination
- `%c` - date and time (%a %b %e %T %Y)
- `%D` - Date (%m/%d/%y)
- `%F` - The ISO 8601 date format (%Y-%m-%d)
- `%v` - VMS date (%e-%^b-%4Y)
- `%x` - Same as %D
- `%X` - Same as %T
- `%r` - 12-hour time (%I:%M:%S %p)
- `%R` - 24-hour time (%H:%M)
- `%T` - 24-hour time (%H:%M:%S)

Jake-Shadle:
This crate https://github.com/MiSawa/time-fmt/ appeared two days ago after I started on this, but it looks to have a few inconsistencies with Ruby/chrono, and doesn't support flags at all, but might be something that could be used in the future.

epage:
Wow, thanks for all of your work on this!

> This crate https://github.com/MiSawa/time-fmt/ appeared two days ago after I started on this, but it looks to have a few inconsistencies with Ruby/chrono, and doesn't support flags at all, but might be something that could be used in the future.

Wonder if they'd accept contributions :)

Jake-Shadle:
I'll link this PR so the author is aware, but I've already spent more time than I would like on this as this was just a side task of purging all uses of chrono from our codebase, and this crate was the last external user of it,

VuiMuich:
Hi, are you intending to bump the crate.io crate after this merge immediatelly?
If so, we would like to delay [leftwm #697](https://github.com/leftwm/leftwm/pull/697) and bump the crate with that PR.

epage:
> Hi, are you intending to bump the crate.io crate after this merge immediatelly? If so, we would like to delay [leftwm #697](https://github.com/leftwm/leftwm/pull/697) and bump the crate with that PR.

Yes, we'll release this immediately.

epage:
> I'll link this PR so the author is aware, but I've already spent more time than I would like on this as this was just a side task of purging all uses of chrono from our codebase, and this crate was the last external user of it,

I can definitely understand.  Thanks for all of your work on this!

epage:
v0.24 is now released

VuiMuich:
Awesome 😎
Thanks for all this!

PR #464, opened 2022-03-29
docs: Correct version references

djwf:
Version reference in lib.rs was outdated.

### DIFF
--- a/crates/bin/README.md
+++ b/crates/bin/README.md
@@ -1,4 +1,4 @@
-liquid-compiler
+liquid-bin
 ===========
 
 > [Liquid templating](http://liquidmarkup.org/) for Rust
@@ -14,7 +14,7 @@ To include liquid in your project add the following to your Cargo.toml:
 
 ```toml
 [dependencies]
-liquid-compiler = "0.20"
+liquid-bin = "0.25.0"
 ```
 
 Now you can use the crate in your code:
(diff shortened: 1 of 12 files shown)


djwf:
As well as a few other spots

crates/bin/README.md
```diff
@@ -14,7 +14,7 @@ To include liquid in your project add the following to your Cargo.toml:
 
 ```toml
 [dependencies]
-liquid-compiler = "0.20"
+liquid-compiler = "0.25.0"
```
epage, line 17: To prevent this from happening again, could you modify each `Cargo.toml` to update this automatically?

To see how, look in `/Cargo.toml` for
```toml
[package.metadata.release]
pre-release-replacements = [
```

You can run `cargo release major --no-publish -vvv` and you should see diffs being shown of the replacements being applied.

If not, its still at least an improvement that these are updated and I can try to remember to update the replacements later.

djwf: Yes, will do - thanks for the guidance!

djwf: Actually - a question: should the `liquid-compiler =` in the `crates/bin/README.md` file be renamed to `liquid-bin =`? The names of all the others match up, but not this one.

djwf:
I pushed the recommended fixes. I also changed `liquid-compiler` to `liquid-bin` in the `crates/bin/README.md`, and made sure all the READMEs had the same basic content.

epage (approved):
Thanks!

Issue #465, opened 2022-03-30
with_time unimplemented


djwf:
As discussed in #252, the `with_time` method is unimplemented (always returns `Nil`).

Discussion about implementing that method, and other fixes required to do so, belong here.

djwf:
It turns out that our DateTime implementation does not handle dates and times without offsets, so we'll need another issue for that.

epage:
See https://github.com/cobalt-org/liquid-rust/issues/48

djwf:
Fixed by #466

Issue #252, opened 2018-11-30
`test_date` is failing for unknown reasons


epage:
Example test
```rust
    assert_eq!(
        v!("May"),
        filters!(date, with_time("2006-05-05 10:00:00"), v!("%B"))
    );
```

djwf:
Looking into this further (even though I've been away from the code for about 4-5 years), it looks like this should be expected. In the two places defined, the function `with_time` returns `Nil`:

https://github.com/cobalt-org/liquid-rust/blob/1a6631d0ce69a4c7c72bf11f7284c853b775a3f1/tests/test_helper.rs#L10-L13

and

https://github.com/cobalt-org/liquid-rust/blob/1a6631d0ce69a4c7c72bf11f7284c853b775a3f1/crates/lib/tests/test_helper.rs#L10-L13

There are a number of issues in the test_date function. I'm going to break them out into separate test functions to make it easier to see which are failing and which aren't.

For this particular issue, even if you remove the `with_time` function call and only pass the date/time string (`"2006-05-05 10:00:00"`), the format by which it is trying to interpret the string, `v!("%B")`, is not valid for the string to be parsed.

I think the following would be a better way to test the month name functionality, but please cross-check me:
``` rust
    assert_eq!(
        v!("May"),
        call_filter!(liquid_lib::stdlib::Date, "May", v!("%B")).unwrap()
    );
```

If the intent is something else, please let me know.

epage:
> Looking into this further (even though I've been away from the code for about 4-5 years), it looks like this should be expected. In the two places defined, the function with_time returns Nil:

Can't believe I missed that `with_time` is not implemented.  That needs fixing

> There are a number of issues in the test_date function. I'm going to break them out into separate test functions to make it easier to see which are failing and which aren't.

Any test in `conformance_ruby` should not be cleaned up; they are a direct port from Shopify's liquid and we want them to line up for easier updating.  For example, see their `test_date`:  https://github.com/Shopify/liquid/blob/master/test/integration/standard_filter_test.rb#L530

djwf:
Got it - I'll leave the tests be, and work on implementing `with_time`. :)

Issue #48, opened 2016-06-30
Add a to_date utility function for the date filter


johannhof:
In https://github.com/cobalt-org/liquid-rust/blob/master/src/filters.rs#L308 we use a custom date format to parse dates for the date filter if it's a string.

The original liquid has a `to_date` function that can take different types of input and parse it. We might even parse different date formats that way  https://github.com/Shopify/liquid/blob/master/lib/liquid/utils.rb#L63

johannhof:
That would also be a good opportunity to add support for the magic `"today"` and `"now"` strings, see https://shopify.github.io/liquid/filters/date/

epage:
If the liquid syntax allows this, I think it'd be nice to read `now` and `today` from the `Context` and fallback to using the current time if they aren't there or aren't compatible.

This would be a bit non-conformant but would be a great for testing.

djwf:
That would be interesting, but I'm not sure how to make that work within the filter.  Conceptually, we'd need to have "today" and "now" be special terms that would be rendered external to the filter if they exist within the Context, and only be handled by the filter if they are not defined within the Context.  Is that what you were thinking?

epage:
For https://github.com/cobalt-org/cobalt.rs/issues/349, I looked up how shopify's date does its parsing:

`date` calls `Utils.to_date` (referenced above by johannhof)
- [`Utils.to_date`](https://github.com/Shopify/liquid/blob/master/lib/liquid/utils.rb#L63)
  - If already a date time, no need to parse
  - If looks like an integer, [parse as timestamp](https://ruby-doc.org/core-2.2.0/Time.html#method-c-at)
  - If looks like a string, `downcase` and use [`Time.parse`](https://svn.ruby-lang.org/cgi-bin/viewvc.cgi/trunk/lib/time.rb?revision=60216&view=markup#l312) which is implemented by [`date__parse`](https://github.com/racker/ruby-1.9.3-lucid/blob/cdb0ad5299aaae6ca14d669738d65bf85691bfef/ext/date/date_parse.c#L1520) ([upstream without syntax highlighting](https://svn.ruby-lang.org/cgi-bin/viewvc.cgi/trunk/ext/date/date_parse.c?revision=56317&view=markup#l2049)).
    - This supports a *lot* of different formats, subsets, etc.

The remaining question is what should serde support do.  Do we support a single format or accept any format?

[...]

PR #466, opened 2022-04-01
fix!: Add formats to DateTimeImpl and implement `with_time` test helper

djwf:

As discussed in #48, #252, and #465 the `DateTimeImpl` needs more formats, and [`with_time`][wt] was unimplemented. These changes add three new formats to `DateTimeImpl`, handle date and time strings (of the supported formats) with no offsets, handle no value (None or the blank string), implements [`with_time`][wt], and splits the ruby conformance [`test_date`][td] function into multiple functions that group the tests logically (and allow isolation of the currently failing tests). The ruby conformance tests are still in the exact same order as in the [original source][os].

[wt]: https://github.com/cobalt-org/liquid-rust/blob/8a2bc481aaea410caa4ef73bd02f23b7655faff7/tests/test_helper.rs#L11
[td]: https://github.com/cobalt-org/liquid-rust/blob/8a2bc481aaea410caa4ef73bd02f23b7655faff7/tests/conformance_ruby/standard_filter_test.rs#L770
[os]: https://github.com/Shopify/liquid/blob/4af38bc/test/integration/standard_filter_test.rb#L529

Feedback welcome!


crates/core/Cargo.toml
```diff
@@ -31,6 +31,7 @@ pest = "2.0"
 pest_derive = "2.0"
 
 # Exposed in API
+regex = "1.5"
```
epage, line 34: This dependency was added in the "public dependencies" section.  I'm not aware of this being a public dependency though

crates/core/src/model/scalar/datetime.rs
```diff
@@ -182,19 +182,52 @@ mod friendly_date_time {
...
+//
+// Offsets in one of the following forms, and are catenated with any of
+// the above formats.
+//
+// * `+HHMM`
+// * `-HHMM`
+//
+// Example:
+//
+// * `dow_mon` format with an offset: "Tue Feb 16 10:00:00 2016 +0100"
```
epage, line 206: Please make this a doc comment, even if the fucntion is private

crates/core/src/model/scalar/datetime.rs
```diff
@@ -182,19 +182,52 @@ mod friendly_date_time {
...
-        "now" => Some(DateTimeImpl::now_utc()),
-        _ => USER_FORMATS
+    if let "" = s {
+        None
+    } else if let "now" | "today" | "Today" = s {
+        Some(DateTimeImpl::now_utc())
+    } else {
+        let offset_re = Regex::new(r"[+-][01][0-9]{3}$").unwrap();
+
+        let s = s.to_owned() + if offset_re.is_match(s) { "" } else { " +0000" };
```
epage, line 226: Please simplify this by breaking the conditional expression out into its own variable
```rust
let offset = if offset_re.is_match(s) { "" } else { " +0000" };
let s = s.to_owned() + offset;
```

crates/core/src/model/scalar/datetime.rs
```diff
@@ -182,19 +182,52 @@ mod friendly_date_time {
...
+        format_description!("[month]/[day]/[year] [hour]:[minute]:[second] [offset_hour sign:mandatory][offset_minute]"),
+        format_description!("[weekday repr:short] [month repr:short] [day padding:none] [hour]:[minute]:[second] [year] [offset_hour sign:mandatory][offset_minute]"),
     ];
 
-    match s {
-        "now" => Some(DateTimeImpl::now_utc()),
-        _ => USER_FORMATS
+    if let "" = s {
+        None
+    } else if let "now" | "today" | "Today" = s {
```
epage, line 221: Why are we handling `today` and `Today` separately rather than [downcasing like the ruby version](https://github.com/Shopify/liquid/blob/master/lib/liquid/utils.rb#L70)?

tests/conformance_ruby/standard_filter_test.rs
```diff
@@ -808,7 +807,10 @@ fn test_date() {
         v!("July"),
         call_filter!(liquid_lib::stdlib::Date, "2006-07-05 10:00:00", "%B").unwrap()
     );
+}
 
+#[test]
+fn test_date_serialized_no_offset() {
```
epage, line 813: Please leave the tests as a single function.

epage (commented):
Thanks for doing this!

btw the PR description looks like its on conventional format.  `!` means "breaking change.  In what way is this expected to be a breaking change?  It'd help to call that out explicitly

crates/core/Cargo.toml
djwf: Whoops, thanks... fixing that now.

djwf:
Sorry! That's my inexperience with the standard format - let me reread the guidelines. :)

djwf:
Ok, I didn't even know that https://www.conventionalcommits.org/en/v1.0.0/ existed, but now I do: thanks! Based on my reading, these aren't actually fixes, they're features. So the commit message would have "feat(core): ...". Does that seem correct?

tests/conformance_ruby/standard_filter_test.rs
djwf: I figured that might be the case, so I made that a separate commit - I've backed it out.

epage:
> Ok, I didn't even know that https://www.conventionalcommits.org/en/v1.0.0/ existed, but now I do: thanks! Based on my reading, these aren't actually fixes, they're features. So the commit message would have "feat(core): ...". Does that seem correct?

Meh, line between bug and feature can be blurry.  For example, we are deviating from user expectations (fix) but we need to add something new to resolve that (feature).

Do what makes sense for communicating intent and don't get too caught up in it.

crates/core/Cargo.toml
djwf: Moved into private area.

crates/core/src/model/scalar/datetime.rs
djwf: Done.

crates/core/src/model/scalar/datetime.rs
djwf: Because I didn't think about it - done.

crates/core/src/model/scalar/datetime.rs
djwf: Whoops - thought I had. Done.

djwf:
There, should be all set now. Let me know if you see anything else that needs fixing!

djwf:
And I just realized it would have been much easier for you had I just pushed changes, rather than force pushing. I'll do that next time. :)

Issue #471, opened 2022-06-15
Render Tag


YuiYukihira:
Hi, sorry if this is a misunderstanding on my part.

But looking [online](https://shopify.github.io/liquid/tags/template/#include), the "include" tag is deprecated and the [render](https://shopify.github.io/liquid/tags/template/#render) tag should be used instead.

I was wondering what the status of this is and if there is any intention to add the new tag? Thanks

epage:
Yes, we need to resolve #419 and close the gap on that.  Unfortunately, a lot of my attention is elsewhere atm so I'm not getting to spend as much time on liquid as I would like

YuiYukihira:
Thanks for confirming this for me. Would you be willing to accept a PR for the render tag if I were to make one?

PR #472, opened 2022-06-22
feat(lib): Add render tag

YuiYukihira:
Hey, here's my PR for #471.

The changes I've made are:
1. I moved the `ForloopObject` and `Range` structs into a parent module, this is because I wanted to use them for the `for` variant of the `render` tag.
2. I've created a new `OpaqueStackFrame` struct, I needed a way to make sure that only the variables passed into the `render` tag could be accessed from inside it.
3. I've then added the `render` tag.

I wasn't sure about whether changes 1 and 2 were acceptable, so let me know what you think.

Thanks


crates/lib/src/stdlib/tags/render_tag.rs
```diff
@@ -0,0 +1,474 @@
...
+use liquid_core::Runtime;
+use liquid_core::ValueView;
+use liquid_core::{Error, Result};
+use liquid_core::{ParseTag, TagReflection, TagTokenIter};
+
+use crate::stdlib::ForloopObject;
+use crate::stdlib::RangeExpression;
+
+#[derive(Copy, Clone, Debug, Default)]
+pub struct RenderTag;
```
epage, line 21: Should we update `ParserBuilder::stdlib` to include this?

crates/core/src/runtime/stack.rs
```diff
@@ -238,3 +238,174 @@ impl<P: super::Runtime> super::Runtime for IndexFrame<P> {
         self.parent.registers()
     }
 }
+
+/// A [`StackFrame`] where variables are not recursively searched for,
+/// However, you can still access the parent's partials.
+pub struct OpaqueStackFrame<P, O> {
```
epage, line 244: nit: I think I'd prefer `SandboxedStackFrame` as I feel like `Opaque` isn't quite getting the point across to me

crates/core/src/runtime/stack.rs
```diff
@@ -238,3 +238,174 @@ impl<P: super::Runtime> super::Runtime for IndexFrame<P> {
...
+        self.parent.set_index(name, val)
+    }
+
+    fn get_index<'a>(&'a self, name: &str) -> Option<ValueCow<'a>> {
+        self.parent.get_index(name)
+    }
+
+    fn registers(&self) -> &super::Registers {
+        self.parent.registers()
+    }
```
epage, line 323: Do these need to be sandboxed as well?

crates/lib/src/stdlib/blocks/for_block.rs
```diff
@@ -225,42 +226,6 @@ impl Renderable for For {
     }
 }
 
-#[derive(Debug, Clone, ValueView, ObjectView)]
-struct ForloopObject<'p> {
```
epage, line 229: I think I'd prefer these staying here and either
- marking them as `pub` and having the `mod.rs` `use` them up the chain
- make the layers of `mod`s `pub` and have the render tag access it

My reason is this is all very tightly part of being a for block and would like to have that code next to each other rather than spread out.   it just happens that `render_tag` wraps the for block behavior so it makes sense to reach in to access it.

crates/lib/src/stdlib/tags/render_tag.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+}
+
+impl TagReflection for RenderTag {
+    fn tag(&self) -> &str {
+        "render"
+    }
+
+    fn description(&self) -> &str {
+        ""
```
epage, line 35: Mind putting one in?

crates/lib/src/stdlib/blocks/for_block.rs
YuiYukihira: Yeah, this make sense. I've gone with the first option as it looks a bit cleaner in my opinion.

crates/core/src/runtime/stack.rs
YuiYukihira: That's fair. I couldn't really think a good name, but sandboxed works.

crates/lib/src/stdlib/tags/render_tag.rs
```diff
@@ -0,0 +1,474 @@
...
+                            .value_with(|| format!("{}", i + 1).into())?;
+
+                        // given that we're at the end of the loop body
+                        // already, dealing with a `continue` signal is just
+                        // clearing the interrupt and carrying on as normal. A
+                        // `break` requires some special handling, though.
+                        let current_interrupt =
+                            scope.registers().get_mut::<InterruptRegister>().reset();
+                        if let Some(Interrupt::Break) = current_interrupt {
+                            break;
```
epage, line 229: Do we need any tests for this?

crates/lib/src/stdlib/tags/render_tag.rs
```diff
@@ -0,0 +1,474 @@
...
+                        >::new();
+                        root.insert("foorloop".into(), &forloop);
+                        root.insert(var_name.as_ref(), &v);
+
+                        let scope = SandboxedStackFrame::new(runtime, &root);
+
+                        let partial = scope
+                            .partials()
+                            .get(&name)
+                            .trace_with(|| format!("{{% render {} %}}", self.partial).into())?;
```
epage, line 214: from [the docs](https://shopify.github.io/liquid/tags/template/#render)

> Note that you don’t need to write the file’s .liquid extension.

Do we need to handle this here?

epage (commented):
I'm sorry for the long delay; somehow I lost track of this.

YuiYukihira:
No worries, I completely forgot about this too.

crates/lib/src/stdlib/tags/render_tag.rs
YuiYukihira: This section is pretty much an exact copy from [here](https://github.com/cobalt-org/liquid-rust/blob/master/crates/lib/src/stdlib/blocks/for_block.rs#L212), in this project. I can't see any specific tests for it there.

epage:
... no idea why my push closed this but it won't let me re-open.  I'll create a new PR

epage:
This is now superseded by #506

Felt bad at how slow I've been at getting back to you so I went ahead and wrapped this up
- Noticed that `,` variables can be used with `with` and `for`, so I adjusted that
- Ported the ruby tests so we know what works and what doesn't.
  - The main thing we don't yet support is `as` being optional.  It should take the file name of the partial to serve as the variable name

PR #473, opened 2022-07-03
Make number coercion behave more like Ruby

nbudin:

liquid-rust doesn't quite behave like the Ruby implementation of Liquid when it comes to coercing other types into integers and floats.  

There are three major differences I've found between Rust's behavior and Ruby's here, and I've tried to account for all of them in this PR:

* Ruby's behavior is to return 0 when converting an unparsable string to an int, and return 0.0 when converting an unparsable string to a float.  Example:
  ```
  irb(main):004:0> Liquid::Template.parse("{{ 'something' | plus: 0 }}").render
  => "0"
  irb(main):005:0> 'something'.to_i
  => 0
  ```
* Ruby truncates floating point numbers when converting to integers.
* Similarly, if asked to convert a string representation of a floating point number to an integer, Ruby will truncate it.

Fixing these behaviors does constitute a breaking change for liquid-rust, since in order to fix them, I had to change several existing tests.  I'm also not really sure that my approach to parsing floating-point number strings as integers is the best one possible - it feels a little hacky and perhaps there's a better way.

### DIFF
--- a/crates/core/src/model/scalar/mod.rs
+++ b/crates/core/src/model/scalar/mod.rs
@@ -99,7 +99,12 @@ impl<'s> ScalarCow<'s> {
     pub fn to_integer(&self) -> Option<i64> {
         match self.0 {
             ScalarCowEnum::Integer(ref x) => Some(*x),
-            ScalarCowEnum::Str(ref x) => x.parse::<i64>().ok(),
+            ScalarCowEnum::Float(ref x) => Some(x.trunc() as i64),
+            ScalarCowEnum::Str(ref x) => Some(
+                x.parse::<f64>()
+                    .map(|parsed| parsed.trunc() as i64)
+                    .unwrap_or(0),
+            ),
             _ => None,
         }
     }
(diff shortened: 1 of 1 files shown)


epage:
Looks like this broke some other cases somehow.

nbudin:
> Looks like this broke some other cases somehow.

Ah yeah, I see that it did.  I'll look into those and see if I can figure out what's up.

nbudin:
@epage First off, sorry for breaking the tests!  This is my first time working with a workspace project, and I didn't realize that simply running `cargo test` without `--workspace` would skip most of the tests.

I've worked through all the test failures.  It appears that a bunch of logic internal to the stdlib as well as the derive macros was assuming that an unparsable integer would come back as None.  In order to keep compatibility with this logic, I've introduced `to_integer_strict` and `to_float_strict` methods that preserve the prior behavior, and changed all the places that seemed to rely on it to use these instead.  The tests are now passing.

Issue #490, opened 2023-01-09
Open to the idea of changing supporting less strict rendering as an option?


RyanAmos:
Following up from my comment left on [this issue](https://github.com/cobalt-org/liquid-rust/issues/299#issuecomment-1374198184).

I have a use-case where it's necessary for us to reduce the rendering strictness, specifically for variables. We allow users to define their own templates and store them. Those templates are then used later to render values with global variables provided by the customers' data which may or may not have all of the variables present. Our current implementation just renders those missing variables as an empty string.

I've tried multiple work arounds to get around this, some of which were fairly hacked together, but they all had edge cases that weren't handled properly.

I'm curious if you're open to the idea of adding a "lax" rendering mode?

epage:
I'd be open to an outside contributor adding this if
- It isn't too invasive
- It doesn't dramatically affect performance
- It is off by default

The first issue I see with implementing this is "where".  The `Runtime` trait is a natural place but we build up runtimes throughout the execution of the template, stacking them on top of each other, so there is no top-level runtime to provide this behavior.

We might be able to handle a templated type that wraps the `Runtime` and all calls actually take this, rather than a runtime directly to ensure it gets used.  It'll get a bit messy but likely would work.

RyanAmos:
PR opened, open to any and all suggestions:
https://github.com/cobalt-org/liquid-rust/pull/492

PR #492, opened 2023-01-10
Lax rendering and parsing

RyanAmos:

Following up on the issue I created here:
https://github.com/cobalt-org/liquid-rust/issues/490

This PR introduces two functional changes. The first being supporting a "lax" parsing mode, the second being a "lax" rendering mode.  Each additional change extends the functionality, but does not change the default behavior. 

### Parsing Mode:
In lax parsing mode, an undefined filter will simply return an internal
noop filter. That filter will simply pass along the string to the next
filter, if one is provided.

In strict parsing mode, an undefined filter will continue to return an
error.

### Rendering Mode:
In lax rendering mode, an undefined variable will simply be rendered as
a `nil` value which will ultimately result in an empty string.

In strict rendering mode, an undefined variable will continue to return
an error.

### DIFF
--- a/crates/core/src/parser/lang.rs
+++ b/crates/core/src/parser/lang.rs
@@ -3,12 +3,25 @@ use super::ParseFilter;
 use super::ParseTag;
 use super::PluginRegistry;
 
+#[derive(Clone)]
+pub enum ParseMode {
+    Strict,
+    Lax,
+}
+
+impl Default for ParseMode {
+    fn default() -> Self {
+        Self::Strict
+    }
+}
+
 #[derive(Clone, Default)]
 #[non_exhaustive]
 pub struct Language {
     pub blocks: PluginRegistry<Box<dyn ParseBlock>>,
     pub tags: PluginRegistry<Box<dyn ParseTag>>,
     pub filters: PluginRegistry<Box<dyn ParseFilter>>,
+    pub mode: ParseMode,
 }
 
 impl Language {
--- a/crates/core/src/runtime/runtime.rs
+++ b/crates/core/src/runtime/runtime.rs
@@ -7,6 +7,14 @@ use crate::model::{Object, ObjectView, Scalar, ScalarCow, Value, ValueCow, Value
 use super::PartialStore;
 use super::Renderable;
 
+/// What mode to use when rendering.
+pub enum RenderingMode {
+    /// Returns an error when a variable is not defined.
+    Strict,
+    /// Replaces missing variables with an empty string.
+    Lax,
+}
+
 /// State for rendering a template
 pub trait Runtime {
     /// Partial templates for inclusion.
@@ -36,6 +44,9 @@ pub trait Runtime {
 
     /// Unnamed state for plugins during rendering
     fn registers(&self) -> &Registers;
+
+    /// Used to set the mode when rendering
+    fn render_mode(&self) -> &RenderingMode;
 }
 
 impl<'r, R: Runtime + ?Sized> Runtime for &'r R {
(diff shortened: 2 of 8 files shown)


src/parser.rs
```diff
@@ -110,6 +112,12 @@ where
             .filter(stdlib::Where)
     }
 
+    /// Sets the parse mode to lax.
+    pub fn in_lax_mode(mut self) -> Self {
+        self.mode = ParseMode::Lax;
+        self
+    }
```
epage, line 119: I think I'd rather re-export the num and accept it.

crates/core/src/parser/lang.rs
```diff
@@ -3,12 +3,25 @@ use super::ParseFilter;
 use super::ParseTag;
 use super::PluginRegistry;
 
+#[derive(Clone)]
+pub enum ParseMode {
+    Strict,
+    Lax,
+}
```
epage, line 10: Also, no docs on this

crates/core/src/parser/lang.rs
```diff
@@ -3,12 +3,25 @@ use super::ParseFilter;
 use super::ParseTag;
 use super::PluginRegistry;
 
+#[derive(Clone)]
+pub enum ParseMode {
```
epage, line 7: I feel like this should also have `Copy`, `Debug`, `PartialEq` and `Eq`

crates/core/src/runtime/runtime.rs
```diff
@@ -7,6 +7,14 @@ use crate::model::{Object, ObjectView, Scalar, ScalarCow, Value, ValueCow, Value
 use super::PartialStore;
 use super::Renderable;
 
+/// What mode to use when rendering.
+pub enum RenderingMode {
```
epage, line 11: I feel like this should have `Copy`, `Clone`, `Debug`, `PartialEq`, `Eq`, and `Default`

src/template.rs
```diff
@@ -14,16 +15,51 @@ pub struct Template {
...
+    /// Renders an instance of the Template, using the given globals.
+    pub fn render_to(&self, writer: &mut dyn Write, globals: &dyn crate::ObjectView) -> Result<()> {
+        self.render_to_with_mode(writer, globals, RenderingMode::Strict)
+    }
+
+    /// Renders an instance of the Template, using the given globals in lax mode.
+    pub fn render_lax(&self, globals: &dyn crate::ObjectView) -> Result<String> {
+        self.render_with_mode(globals, RenderingMode::Lax)
+    }
+
```
epage, line 30: Rather than having a bunch of different render funtions, should this be stored with the Parser and passed in when this is created?

crates/core/src/parser/lang.rs
```diff
@@ -3,12 +3,25 @@ use super::ParseFilter;
...
+    Strict,
+    Lax,
+}
+
+impl Default for ParseMode {
+    fn default() -> Self {
+        Self::Strict
+    }
+}
```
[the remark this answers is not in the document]
RyanAmos: > This is a very general name that only covers filters. The name could imply tags, blocks, syntax errors as well.
> 
> Should we make this more specific?

Fair.  I can update the name to be more specific.

> Also, the line for what is parse vs render might be a bit fuzzy for a user that limiting these to just Render and Parse modes obscures the behavior.

Can you elaborate a bit more what you mean here?

src/parser.rs
RyanAmos: I'll update, I did it this way since there's only 2 options currently, 1 of which is the default.  So exposing a function allowed us to now have to worry about exposing the enum to the user.

crates/core/src/parser/lang.rs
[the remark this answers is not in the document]
epage: > Can you elaborate a bit more what you mean here?

Is it required that filter lookups happen during parse or could it happen during runtime?  At the moment, that is an implementation detail that can change.

src/template.rs
RyanAmos: I actually originally started that route but decided against it.

When stored on the parser that means you're unable to determine which mode you'd want to use after the parser is created.  Every template that's parsed would be stuck in that rendering mode.  People would in theory, need to then create two separate parsers to work around that limitation.  Admittedly, I'm not sure how frequent people would ever need to change the rendering mode of a parser. I don't personally have a use case to argue that.

Another option is we could expose a function to set the rendering mode on the `Template` instead.  I'm not convinced that's a better UX, but probably a bit less confusing to most users who may use this solely in "strict" mode.

crates/core/src/parser/lang.rs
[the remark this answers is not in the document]
RyanAmos: You're right that the implementation can change, but I personally think the way it's currently implemented is the correct behavior.

The way it's currently implemented allows you to store a template that are parsed once and reused multiple times.  "Lazily" parsing the template means you delay errors that IMO should be handled when the template itself is created and in turn parsed.

Happy to explore alternative options if you think that's the right approach.

PR #495, opened 2023-03-28
fix: DateFilter doesn't support date format

flisky:


flisky:
actually, there is no `datetime` filter, and it seems that `Date` variant should be replaced by `DateTime`?

PR #496, opened 2023-04-08
Allow filter chains to start with a missing variable

### LINKED ISSUE #477: EmptyDrop and Object indexing
liquid-rust version: 0.26.0
rust version: 1.63.0
OS: linux

I've been referencing [this page](https://shopify.github.io/liquid/basics/types/) of the liquid docs, and my understanding of the `nil` and `EmptyDrop` type is that both of the following should be possible:

```liquid
{% assign var = value | default: 42 %}
{% assign var = data.value | default: 42 %}
```

In using liquid-rust, however, I'm finding that the second case fails for me if `data.value` does not already exist. I get an error like:

```
liquid: Unknown index
  with:
    variable=data
    requested index=value
    available indexes=other, more_data
from: {% assign var = data["value"] | default : 42%}
```

I was looking for an existing issue for this but didn't find anything, am I missing something?

rydrman:

Following up on #477, this change treats missing variables and indexes as `nil` when they appear at the start of a filter chain. As far as I can tell, this is in line with the [shopify implementation](https://shopify.dev/docs/api/liquid/filters/default).

I needed a way to properly identify when an error identified a missing index or variable, and so introduced the concept of an `ErrorKind` to handle this. If there is a better or preferred way, please let me know! 

closes #477

### DIFF
--- a/crates/core/src/error/error.rs
+++ b/crates/core/src/error/error.rs
@@ -16,26 +16,50 @@ pub struct Error {
     inner: Box<InnerError>,
 }
 
+impl Error {
+    /// Identifies the underlying kind for this error
+    pub fn kind(&self) -> &ErrorKind {
+        &self.inner.kind
+    }
+}
+
 // Guts of `Error` here to keep `Error`'s memory size small to avoid bloating the size of
 // `Result<T>` in the success case and spilling over from register-based returns to stack-based
 // returns.  There are already enough memory allocations below, one more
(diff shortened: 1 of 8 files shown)


jesseditson:
I'd love to see this land, but it looks pretty stale - reading https://github.com/cobalt-org/liquid-rust/issues/477 it's unclear to me - is the resolution that this should or should not be the correct behavior?

PR #500, opened 2023-05-27
fix: Correct date time serde formats

tglman:
Hi,

Since this [PR](https://github.com/cobalt-org/liquid-rust/pull/494) the date format in serde require milliseconds, this break the downstream project cobalt, that if updated require the dates in the frontmatter to have the milliseconds in the published_date, this PR correct that problem having the support of parsing both dates with and without milliseconds, and optionally include the milliseconds in serialization in case are a value != 0.

### DIFF
--- a/crates/core/src/model/scalar/datetime.rs
+++ b/crates/core/src/model/scalar/datetime.rs
@@ -174,8 +174,13 @@ mod friendly_date_time {
     where
         S: Serializer,
     {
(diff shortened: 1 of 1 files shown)


epage:
Thanks for doing this!

For context for anyone else, we already do this in the `Display`, I figured it would be good to mirror this in our serde support.

Issue #507, opened 2023-06-13
Question: supporting object literals with unknown keys on Custom filters


juanazam:
Hello!

I'm working on supporting Shopify's `t` filter (translation filter).

Here there is a simple a example of how it works. If we have the following translations file

```
/locales/en.default.json

{
  "layout": {
    "header": {
      "hello_user": "Hello {{ name }}!"
    }
  }
}
```

And the following liquid template:

```
<h1>{{ 'layout.header.hello_user' | t: name: customer.first_name }}</h1>
```

We would expect this example to render (assuming the first name of the customer is John):

```
<h1>Hello John</h1>
```

I'm running into the following issue when implementing this as a custom filter.
This is how I'm defining the filter's parameters:

```rust
#[derive(Debug, FilterParameters)]
struct TArgs {
    #[parameter(
        description = "Variables to be used for revaluating liquid once translation is resolved.",
    )]
    variables: Option<Expression>,
}

#[derive(Clone, ParseFilter, FilterReflection)]
#[filter(
    name = "t",
    description = "Access the desired translation for the recipient local",
    parameters(TArgs),
    parsed(TFilter)
)]
pub struct T;

#[derive(Debug, FromFilterParameters, Display_filter)]
#[name = "t"]
struct TFilter {
    #[parameters]
    args: TArgs,
}
```

But the problem is that when I run the same example I get the following error:

```
 "liquid: Unexpected named argument `name`\nfrom: Filter parsing error\n  with:\n    filter=t: name: customer.name\n"}
 ```
 
I haven't been able to get the filter working with object literals with arbitrary keys. What's the best way to accomplish this?

juanazam:
Hi @epage, sorry to bother you, do you have any insights on how I might be able to achieve this?
Tagging you directly because you seem to be the person with the most knowledge con this crate, thanks in advance!

TomzBench:
I made a custom `json` filter by importing serde.  Since the liquid model implements `Serialize/deserialize` you can use it to get a `serde_json::Value` and loop through keys...

```rs
use liquid_core::{Display_filter, Filter, FilterReflection, ParseFilter};
use liquid_core::{Error, Result, Runtime};
use liquid_core::{Value, ValueView};

#[derive(Clone, ParseFilter, FilterReflection)]
#[filter(
    name = "json",
    description = "Convert a JSON string into a liquid object",
    parsed(JsonFilter)
)]
pub struct Json;

[...]

PR #516, opened 2023-09-18
feat: support keyword arguments with no definition on custom filters

juanazam:

Implements the changes required to support what's described on #507.
In order to support arbitrary keyword arguments, this PR introduces a new type of parameter.

When defined, the `keyword_group` parameter is populated with all keyword parameters used to call the filter.
It also disables the validation of only allowing known keyword parameters to be used on custom filters. 

It also introduces a new variant for Expression (defined at runtime) to support this.

I can update this pr in any way maintainers feel it makes more sense.


epage:
This feels very hacked-in.  Any idea how this is implemented with the ruby implementation?  Its been a while since I've looked but my guess is they expose raw text for it to parse as they want rather than having any kind of higher level constructs to work with.

juanazam:
Hi @edpage

Thanks for the feedback.
The ruby implementation is quite simple, because given it’s dynamic, they translation filter receives the translation key and the variables as a hash.

Here is the implementation: https://github.com/Shopify/liquid/blob/0b9318222bcc09681e52fd5b8e70262274e673bf/lib/liquid/i18n.rb#L17
And here an example on a test: https://github.com/Shopify/liquid/blob/0b9318222bcc09681e52fd5b8e70262274e673bf/test/unit/i18n_unit_test.rb#L21
The fixture file is here: https://github.com/Shopify/liquid/blob/0b9318222bcc09681e52fd5b8e70262274e673bf/test/fixtures/en_locale.yml

The translation fixture file contains: 

```
---
  simple: "less is more"
  whatever: "something %{something}"
```

The call on the test `@i18n.translate("whatever", something: "different”)` will replace the usage of something with the something variables passed as argument on the filter.

In the case of rust-liquid, I couldn’t find another workaround on how to indicate that I want my custom filter to receive any arbitrary variables and not have to define them
when defining the filter. This is necessary because liquid users can pass any variables to the filter and we cannot know them in advance.

I will be glad to update this PR in any direction that feels better, but I’m not sure what’s the best way besides the one that I found.

juanazam:
Hey @epage any comments/advice on what can be done here?

juanazam:
Hi @epage, any suggestions here on how to handle this?

PR #527, opened 2023-12-22
Add support for indexmap

hardliner66:
I added the possibility to enable the use of an `IndexMap` inside of the templates. I currently use liquid in a tool of mine, where I use an `IndexMap` to keep elements in the order they were inserted in, as re-ordering those could lead to bugs further down.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -24,6 +24,7 @@ num-traits = "0.2"
 pest = "2.0"
 pest_derive = "2.0"
 regex = "1.5"
+indexmap = { version = "2.1.0", optional = true }
 
 # Exposed in API
 time = { version = "0.3", default-features = false, features = ["formatting", "macros", "parsing"] }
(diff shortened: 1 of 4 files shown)


crates/core/Cargo.toml
```diff
@@ -38,3 +39,4 @@ snapbox = "0.4.14"
 [features]
 default = []
 derive = ["liquid-derive"]
+indexmap = ["dep:indexmap"]
```
epage, line 42: Should suffix the feature with a `2` to track that this feature is for v2 (and allow us to support a v3, etc)?

crates/core/src/model/object/mod.rs
```diff
@@ -358,6 +361,79 @@ impl<'s, O: ObjectView> fmt::Display for ObjectRender<'s, O> {
     }
 }
 
+#[cfg(feature = "indexmap")]
+impl<K: ObjectIndex, V: ValueView> ValueView for IndexMap<K, V> {
```
epage, line 365: Could you put this right after `BTreeMap`s impl?

Issue #509, opened 2023-07-01
How do you use `{% include x.liquid %}`?


angeldollface:
liquid-rust version: 0.26.4
rust version: 1.70.0
OS: Mac OSX 13.2.1

I need this functionality for my static-site generator, [Mandy](https://github.com/angeldollface/mandy). 
I have this context structure [here](https://github.com/angeldollface/mandy/blob/main/src/modules/context.rs) which is applied through the `object` macro. How would I handle `include` statements in templates through this? I couldn't find anything in the docs.

epage:
Why would this be marked as closed?  You requested documentation updates which someone is doing in #542

angeldollface:
> Why would this be marked as closed? You requested documentation updates which someone is doing in #542

@epage Wasn't quite sure. Right you are! <3

PR #542, opened 2024-05-09
Add documentation snippet for using partials

santoso-wijaya:
- [N/A] Tests created for any new feature or regression tests for bugfixes.

#509

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,9 +25,9 @@
 mod parser;
 mod template;
 
+pub mod partials;
 pub mod reflection;
 
-pub use liquid_core::partials;
 /// Liquid data model.
 pub mod model {
     pub use liquid_core::array;
--- a/src/partials.rs
+++ b/src/partials.rs
@@ -1,1 +1,44 @@
+//! ## Using partials
+//!
+//! To use `{% include %}` or `{% render %}` tags in a template, you first need to compile these
+//! included files as template partials.
+//!
+//! Example:
+//!
+//! ```liquid
+//! # common.liquid
+//! Number: {{ i }}
+//! ```
+//!
+//! ```rust
+//! use liquid::ParserBuilder;
+//! use liquid::partials::{EagerCompiler, InMemorySource, PartialSource};
+//!
+//! // Build template partials using an eager, in-memory source compiler.
+//! // Other compilation policies also exist depending on specific needs.
+//! type Partials = EagerCompiler<InMemorySource>;
+//!
+//! let partials = {
+//!   let mut partials = Partials::empty();
+//!
+//!   let filepath = String::from("common.liquid");
+//!   //let contents = std::fs::read_to_string(&filepath).unwrap();
+//!   let contents = "Number: {{ i }}";
+//!
+//!   partials.add(filepath, contents);
+//!   partials
+//! };
+//!
+//! // Compile and render the main template, which uses the "common" partial.
+//! let parser = ParserBuilder::with_stdlib().partials(partials).build().unwrap();
+//! let rendered = {
+//!     let mut globals = liquid::object!({ "num": 42 });
+//!     parser
+//!         .parse("Liquid! {% render \"common\", i: num %}").unwrap()
+//!         .render(&globals).unwrap()
+//! };
+//!
+//! assert_eq!(rendered, "Liquid! Number: 42");
+//! ```
+
 pub use liquid_core::partials::*;
--- a/src/partials.rs
+++ b/src/partials.rs
@@ -1,0 +1,1 @@
+pub use liquid_core::partials::*;


README.md
```diff
@@ -92,3 +92,35 @@ See
...
+as template partials.
+
+Example:
+
+```liquid
+# common.liquid
+Number: {{ i }}
+```
+
+```rust
```
epage, line 108: imo this should be a module level comment on `partials`.
- Focuses the documentation in one place
- Verifies the examples

santoso-wijaya: Thanks for the quick review! I moved the documentation as a mod rustdoc as you suggested.

crates/core/src/partials/mod.rs
```diff
@@ -1,3 +1,36 @@
...
+//!
+//! ```liquid
+//! # common.liquid
+//! Number: {{ i }}
+//! ```
+//!
+//! ```rust,no_run,compile_fail
+//! // Build template partials using an eager, in-memory source compiler.
+//! // Other compilation policies also exist depending on specific needs.
+//! type Partials = EagerCompiler<InMemorySource>;
```
santoso-wijaya, line 15: These types (along with ParserBuilder, Parser, etc) are defined outside of this crate in a reverse dependent way, and I can't figure out what the right `# use ...` directive to use to make this snippet compile.

crates/core/src/partials/mod.rs
```diff
@@ -1,3 +1,36 @@
...
+//!
+//! let partials = {
+//!   let mut _partials = Partials::empty();
+//!
+//!   let filepath = String::from("common.liquid");
+//!   let contents = std::fs::read_to_string(filepath).unwrap();
```
santoso-wijaya, line 21: What to do with file reads in doctests

crates/core/src/partials/mod.rs
```diff
@@ -1,3 +1,36 @@
...
+//! Number: {{ i }}
+//! ```
+//!
+//! ```rust,no_run,compile_fail
+//! // Build template partials using an eager, in-memory source compiler.
+//! // Other compilation policies also exist depending on specific needs.
+//! type Partials = EagerCompiler<InMemorySource>;
+//!
+//! let partials = {
+//!   let mut _partials = Partials::empty();
```
epage, line 18: Could you remove the `_` prefix?  That tells the compiler that it can be unused.  We can just call it `partials`

santoso-wijaya: Done.

crates/core/src/partials/mod.rs
```diff
@@ -1,3 +1,36 @@
...
+//! included files as template partials.
+//!
+//! Example:
+//!
+//! ```liquid
+//! # common.liquid
+//! Number: {{ i }}
+//! ```
+//!
+//! ```rust,no_run,compile_fail
```
epage, line 12: Sorry, I had overlooked your comments that you had resolved.

I'm fine with `no_run` examples but I really dislike `ignore` (which is more of what this should be than `compile_fail`).

Instead, let's stop re-exporting `partials` in `liquid` and instead have a dedicated module that manually re-exports the types from `liquid_core::partials`.

What would be helpful is if we organized the PR as the following commits
- A commit to only re-export `partials` types and not the `mod` itself
- A commit that adds the documentation.

As those are the two core atomic operations for this PR.

santoso-wijaya: Got it.

There is still an issue with `ParserBuilder` being defined in in a layer above `liquid_core::partials` here, though.

So, in the rustdoc code that we include here in `crates/core/src/partials/mod.rs`, we would be showing how a user would create a partial to be included in a `liquid::ParserBuilder`, but `liquid::ParserBuilder` being a client of `liquid_core::partials` will not be in scope here to be imported in the rustdoc code.

In other words, this compiles just fine:

```rust

//! ```rust,no_run
//! use liquid_core::partials::{EagerCompiler, InMemorySource, PartialSource};

```

But the following will not:

```rust

//! ```rust,no_run
//! use liquid_core::partials::{EagerCompiler, InMemorySource, PartialSource};
//! use liquid::ParserBuilder;

```

epage: No, I was saying we should create `src/partials.rs` and have the example on that.

santoso-wijaya: Creating `src/partials.rs` makes sense to me. The rustdoc example that we can write in that module won't be able to include and compile this usage snippet though:

```rust
//! // Compile and render the main template, which uses the "common" partial.
//! let parser = ParserBuilder::with_stdlib().partials(partials).build().unwrap();
```

Since `liquid::ParserBuilder` depends on `partials`, not the other way around.

santoso-wijaya: Wait, nevermind, I think that'll work. Lemme prep the commits.

santoso-wijaya: Wait, nevermind, I think that'll work. I'll prep the commits.

santoso-wijaya: Done.

epage:
Thank you for your work and our patience on the feedback!

PR #546, opened 2024-05-31
feat(lib): Implement Jekyll's version of sort

emmyoh:

Part of #311. Continuation of #438.

### DIFF
--- a/crates/core/src/parser/parser.rs
+++ b/crates/core/src/parser/parser.rs
@@ -78,6 +78,16 @@ pub fn parse(text: &str, options: &Language) -> Result<Vec<Box<dyn Renderable>>>
     Ok(renderables)
 }
 
+/// Given a `Variable` as a string, parses it into a `Variable`.
+pub fn parse_variable_from_text(text: &str) -> Result<Variable> {
+    let variable = LiquidParser::parse(Rule::Variable, text)
+        .map_err(convert_pest_error)?
+        .next()
+        .expect("Parsing a variable failed.");
+
+    Ok(parse_variable(variable))
+}
+
 /// Parses a `Scalar` from a `Pair` with a literal value.
 /// This `Pair` must be `Rule::Literal`.
 fn parse_literal(literal: Pair) -> Value {
--- a/crates/core/src/parser/parser.rs
+++ b/crates/core/src/parser/parser.rs
@@ -79,13 +79,13 @@ pub fn parse(text: &str, options: &Language) -> Result<Vec<Box<dyn Renderable>>>
 }
 
 /// Given a `Variable` as a string, parses it into a `Variable`.
-pub fn parse_variable_from_text(text: &str) -> Result<Variable> {
+pub fn parse_variable(text: &str) -> Result<Variable> {
     let variable = LiquidParser::parse(Rule::Variable, text)
         .map_err(convert_pest_error)?
         .next()
         .expect("Parsing a variable failed.");
 
-    Ok(parse_variable(variable))
+    Ok(parse_variable_pair(variable))
 }
 
 /// Parses a `Scalar` from a `Pair` with a literal value.
@@ -134,7 +134,7 @@ fn parse_literal(literal: Pair) -> Value {
 
 /// Parses a `Variable` from a `Pair` with a variable.
 /// This `Pair` must be `Rule::Variable`.
-fn parse_variable(variable: Pair) -> Variable {
+fn parse_variable_pair(variable: Pair) -> Variable {
     if variable.as_rule() != Rule::Variable {
         panic!("Expected variable.");
     }
@@ -173,7 +173,7 @@ fn parse_value(value: Pair) -> Expression {
 
     match value.as_rule() {
         Rule::Literal => Expression::Literal(parse_literal(value)),
-        Rule::Variable => Expression::Variable(parse_variable(value)),
+        Rule::Variable => Expression::Variable(parse_variable_pair(value)),
         _ => unreachable!(),
     }
 }
@@ -981,7 +981,7 @@ impl<'a> TagToken<'a> {
     /// Tries to obtain a `Variable` from this token.
     pub fn expect_variable(mut self) -> TryMatchToken<'a, Variable> {
         match self.unwrap_variable() {
-            Ok(t) => TryMatchToken::Matches(parse_variable(t)),
+            Ok(t) => TryMatchToken::Matches(parse_variable_pair(t)),
             Err(_) => {
                 self.expected.push(Rule::Variable);
                 TryMatchToken::Fails(self)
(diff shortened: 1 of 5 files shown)


crates/lib/src/jekyll/array.rs
```diff
@@ -10,6 +13,122 @@ use liquid_core::{Value, ValueView};
...
+    args: SortArgs,
+}
+
+enum NilsOrder {
+    First,
+    Last,
+}
+
+fn safe_property_getter<'a>(value: &'a Value, property: String) -> Value {
+    let properties = property.split('.').collect::<Vec<&str>>();
```
emmyoh, line 49: Is there a better way to do this? This won't work when the `object["property"]` syntax is used instead of `object.property`, right?

crates/lib/tests/conformance_jekyll/test_filters.rs
emmyoh, line 1: [These tests](https://github.com/jekyll/jekyll/blob/1484c6d6a41196dcaa25daca9ed1f8c32083ff10/test/test_filters.rb#L1383-L1434) are from the Jekyll implementation.

epage:
I was trying to better understand some of what you were doing so I re-wrote your PR in mine, see #438.  I didn't overwrite yours because there ended up being some differences I wanted to check with you on.

crates/lib/src/jekyll/array.rs
```diff
@@ -10,6 +13,122 @@ use liquid_core::{Value, ValueView};
...
+    let mut result = value.to_owned();
+    for property in properties {
+        result = result
+        .as_object()
+        .and_then(move |obj| {
+            if let Some(scalar) = obj.as_scalar() {
+                try_find(&property, &vec![scalar]).map(|v| v.clone().to_value())
```
epage, line 56: Why are we searching within a property to find the value within the array?

My copy doesn't have this and it passes tests.

In fact, why are we manually looping over `properties` rather than calling `try_find(value, properties)`?

emmyoh: The short answer is that I'm not super familiar with the inner workings of this crate. I'll fiddle with this a bit more and update my PR. Thanks for reviewing!
IIRC your implementation did not pass every test (didn't work for nested properties) but I could be misremembering.

epage: I meant my updated copy that I referenced in https://github.com/cobalt-org/liquid-rust/pull/546#issuecomment-2143010012

emmyoh: Re-read your comment and I see what you mean now. I'll fix this when I get the chance.

emmyoh:
@epage Incorporated your changes. Instead of doing text manipulation (eg, splitting on `.` like before) I'm parsing using the grammar itself, which meant adding `parse_variable_from_text` to the API. Understand if you're uncomfortable with doing this; would appreciate further input on how to proceed.

crates/core/src/parser/parser.rs
```diff
@@ -78,6 +78,16 @@ pub fn parse(text: &str, options: &Language) -> Result<Vec<Box<dyn Renderable>>>
...
+/// Given a `Variable` as a string, parses it into a `Variable`.
+pub fn parse_variable_from_text(text: &str) -> Result<Variable> {
+    let variable = LiquidParser::parse(Rule::Variable, text)
+        .map_err(convert_pest_error)?
+        .next()
+        .expect("Parsing a variable failed.");
+
+    Ok(parse_variable(variable))
+}
+
```
emmyoh, line 90: Understand if you're uncomfortable introducing a new public function here. If you take issue with this, what would you recommend?

crates/lib/src/jekyll/array.rs
```diff
@@ -195,6 +323,58 @@ impl Filter for ArrayToSentenceStringFilter {
...
+            }
+        ]);
+        assert_eq!(
+            liquid_core::call_filter!(Sort, input, "date.year").unwrap(),
+            desired_result
+        );
+        assert_eq!(
+            liquid_core::call_filter!(Sort, input, "date[\"year\"]").unwrap(),
+            desired_result
+        );
```
emmyoh, line 375: Ensuring both `object.property` and `object["property"]` work. I could've tested for `object['property']` too, I suppose.

epage:
> @epage Incorporated your changes. Instead of doing text manipulation (eg, splitting on . like before) I'm parsing using the grammar itself, which meant adding parse_variable_from_text to the API. Understand if you're uncomfortable with doing this; would appreciate further input on how to proceed.

My PR was constructed in a fairly particular way.  Could you preserve the commits from that PR, editing them or adding to them as needed?

crates/core/src/parser/parser.rs
```diff
@@ -78,6 +78,16 @@ pub fn parse(text: &str, options: &Language) -> Result<Vec<Box<dyn Renderable>>>
     Ok(renderables)
 }
 
+/// Given a `Variable` as a string, parses it into a `Variable`.
+pub fn parse_variable_from_text(text: &str) -> Result<Variable> {
```
epage, line 82: Let's remove the `_from_text`

crates/core/src/parser/parser.rs
epage: We need to re-do the parsing which will be a breaking change anyways, its likely fine to expose this for now

crates/core/src/parser/parser.rs
epage: (that requires renaming `parse_variable` which I'm fine with, no strong preference on that name)

emmyoh: Sounds good! I'll take care of this later today, and I'll also learn some Git magic to try and merge our PRs together. Thanks for your help!

emmyoh:
> > @epage Incorporated your changes. Instead of doing text manipulation (eg, splitting on . like before) I'm parsing using the grammar itself, which meant adding parse_variable_from_text to the API. Understand if you're uncomfortable with doing this; would appreciate further input on how to proceed.
> 
> 
> 
> My PR was constructed in a fairly particular way.  Could you preserve the commits from that PR, editing them or adding to them as needed?

Does Git have the ability to copy commits into another branch's history? Would creating a merge commit be acceptable? Sorry, not super familiar with all of Git's features. I'll look into this when I get the chance tonight.

epage:
> Does Git have the ability to copy commits into another branch's history? Would creating a merge commit be acceptable? Sorry, not super familiar with all of Git's features. I'll look into this when I get the chance tonight.

`git cherry-pick <sha>`.

You can use that to slowly rebuild this to what you want.  For example, I might do something like
```console
$ git switch jekyll-sort
$ git switch -c old-jekyll-sort
$ git switch jekyll-sort
$ git reset --hard master
$ git cherry-pick c9a61bc215dc1bdd7f664512ebf6f9da46c5fa45 # from #438 
$ ... edit as needed and `git commit --amend`
$ ... loop over, cherry picking each commit
```

emmyoh:
@epage Tried my best to merge the two commit histories together while preserving the order in which changes were made. Excited to get this PR merged! This has been bothering me for a while (three years at this point?), haha.

epage:
Since I was wanting to keep the structure of #438, I went ahead and pulled your changes over into that.

PR #561, opened 2024-09-22
Updated the link to Mandy.

alyxshang:
Hi!

I created Mandy many months ago and decided to rewrite the entire thing from scratch to be

- faster and lighter
- easier to maintain
- work on all major platforms
- have a significantly lighter codebase

The new version of Mandy uses the crate for Liquid but compared to the previous version, Mandy shrunk from around 4300 lines to around 1700 lines. Also went through a rebranding which is why I have a new name and new handle now. 

@epage If you see this, you will remember (probably).


alyxshang:
@epage Thank you. :heart:

